PR #34, opened 2014-12-13
Support for "SEC" compressed keys.

samrushing:
For more info see:
 http://blog.richardkiss.com/?p=371
 http://www.secg.org/
 http://www.secg.org/sec1-v2.pdf


samrushing:
Update: removed the redundant modular_sqrt fun.

tomato42:
@samrushing I'm guessing that you consider this PR to be obsoleted by #54, is that correct?

tomato42:
fixed by #118

PR #37, opened 2015-01-16
change util.randrange to bit-wise

pik:
The current `util.randrange` runs up to worst case mean of 256 sys calls. This is, I think, the simplest way to implement bit-wise instead of byte-wise randrange. In terms of speed bit-wise is a lot more consistent (only where order is close to but less than (256**j) (It should be +2 but the original implementation calculated orderlen incorrectly) does it fall behind). 

``````
>>> timeit(randrange_old, order=generator_521.order(), n=10**5)
(0.0012479020285606384, 128.32433)  #mean time per run, mean os.urandom calls
>>> timeit(randrange_bit, order=generator_521.order(), n=10**5)
(6.894484043121337e-05, 1.0)   #mean time per run, mean os.urandom calls```
``````

### DIFF
--- a/ecdsa/util.py
+++ b/ecdsa/util.py
@@ -3,6 +3,7 @@ from __future__ import division
 import os
 import math
 import binascii
+import sys
 from hashlib import sha256
 from . import der
 from .curves import orderlen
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ecdsa/util.py [hunk 1]
```diff
@@ -16,15 +17,19 @@
 oid_ecPublicKey = (1, 2, 840, 10045, 2, 1)
 encoded_oid_ecPublicKey = der.encode_oid(*oid_ecPublicKey)
 
+if sys.version > '3': 
+    entropy_to_bits = lambda ent_256: ''.join(bin(x)[2:].zfill(8) for x in ent_256)
+else: 
+    entropy_to_bits = lambda ent_256: ''.join(bin(ord(x))[2:].zfill(8) for x in ent_256)
+if sys.version < '2.7':  #Can't add a method to a built-in type so we are stuck with this
+    bit_length = lambda x: len(bin(2)) - 2
```
warner: did you mean `len(bin(x))` here?

pik:
@warner 
You're right, I apologize, that was very careless.

PR #42, opened 2015-03-11
Add from_signature constructor to VerifyingKey

ecdsa:
This constructor creates a VerifyingKey from a signature.
It is currently implemented separately in Electrum, see:
https://github.com/spesmilo/electrum/blob/master/lib/bitcoin.py#L468
I guess it would make sense to add it to python-ecdsa instead


tomato42:
this is now implemented as `from_public_key_recovery` and `from_public_key_recovery_with_digest` in `ecdsa.keys.VerifyingKey`

PR #45, opened 2015-08-09
Bump six version and minor improvements

Elizafox:
Bump the six version, and add Python 2 Unicode support using the `@six.python_2_unicode_compatible` class decorator for CurveFp.

### DIFF
--- a/ecdsa/six.py
+++ b/ecdsa/six.py
@@ -1,33 +1,39 @@
 """Utilities for writing code that runs on Python 2 and 3"""
 
-# Copyright (c) 2010-2012 Benjamin Peterson
+# Copyright (c) 2010-2015 Benjamin Peterson
 #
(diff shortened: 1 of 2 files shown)


ecdsa/six.py
```diff
@@ -61,7 +67,7 @@ def __len__(self):
         else:
             # 64-bit
             MAXSIZE = int((1 << 63) - 1)
-            del X
+        del X
```
tomato42: I think you meant here
```
finally:
    del X
```

warner:
I rebased this, updated six to the most recent version (1.10.0), made some more fixes, and landed the result in 04a383c7bc144453593192005b78aba2e989c91e . Thanks!

@tomato42 the change you suggest makes sense to me.. I think it needs to go upstream into six instead. Could you file an issue at https://bitbucket.org/gutworth/six ?

tomato42:
don't have a bitbucket account

PR #47, opened 2015-09-28
Fix parameter call of openssl.

cornelinux:
The parameters with openssl are using only one dash.


warner:
landed (rebased) in 19eb97d3074510be4a3ef2b34133fdf34d79d56e, thanks!

PR #48, opened 2015-10-10
Make Python 3.5 support official

tomato42:
Since the test suite works on Python 3.5 without any problems, make the support official. Also make sure that it stays like this by including it in Travis setup


warner:
landed (rebased) in fbcc710f526c479bba23185a6e968ee3f247b470, thanks!

PR #49, opened 2015-10-10
fix module name in examples

tomato42:
The module naming in README is inconsistent, once referring to `pyecdsa` while other times using `ecdsa`.

Change the documentation to use `ecdsa` throughout.


warner:
landed (rebased) in 4db8244e2ec9b86cb6347027e0f0b070ae4221d1, thanks!

Issue #53, opened 2016-01-25
Incompatibility


trancee:
I am trying to do message signature authentication over different systems, e.g. JavaScript (http://kjur.github.io/jsrsasign/sample-ecdsa.html) with python-ecdsa. OpenSSL is the baseline for the tests and the JavaScript library is very close to produce the same results. I was expecting the same from python-ecdsa. Unless I am doing something wrong, the public key is always different in size and thus the signature validation fails.

Those are the steps I am doing:
1.) First I am deciding which curve to use, in this case it is **prime256v1**.
2.) Then I generate the private and public keys in Python by doing this:
>`    privateKey = SigningKey.generate(curve=NIST256p)`
>`    publicKey = privateKey.get_verifying_key()`

3.) That results in a key pair which the hex representation looks like this:
>`    privateKey = d1d801d3166dc5b7acc2651fc1b901a6f5fc8542851a505e7b76af4c4302a85a`
>`    publicKey = 88d1b89b5e1f1a7eeb2007c31416ce743b4e3c23e261ca8b325e746938b218baec1c9d6120ffe91fe8a93d81cc55aa1cb4c995886dba91efd82ac4497cd9abf9`

4.) Applying the key pair to the JavaScript testing environment, I can sign the message, but when verifying it, it says the signature is invalid.

When I generate the keys using the JavaScript implementation, I get the following as an example:
>`    privateKey = 7e4cc8d77d6109ac01f47d530945d828673e0809ba3953b14f6fc7558ab6f670`
>`    publicKey = 0499dc501f2fe17cc8512a689498d7e1dd4c717d78db8d6a8b40c4bfeed50c01f6120ab2d09eccf741baf04f8f1451bcc99be8d48d0769837d18be934471dc1a1e`

As you can see, the privateKey's are both of same size, but the publicKey's are different in size.

To determine if the problem is the JavaScript implementation, I followed the same steps using OpenSSL:

`openssl ecparam -genkey -name prime256v1 -out k.pem`
`openssl ec -in k.pem -noout -text`

>`Private-Key: (256 bit)`
>`priv:`
>`    60:54:0c:84:8e:ec:de:b0:df:e6:2e:02:96:f3:d3:`
>`    3b:40:e3:fe:14:e8:f8:88:97:5c:bd:a3:2c:3e:5a:`
>`    cb:57`
>`pub: `
>`    04:c4:9a:71:9d:93:fe:7a:24:c1:fd:ce:c2:28:6a:`
[...]

PR #54, opened 2016-02-07
Support for SEC compressed keys

### LINKED ISSUE #1: point compression: reduce pubkey size in half
Point-compression is a well-understood technique for reducing the size of ECDSA public keys in half, making them the same size as the private key (plus a single sign bit). The signatures are still twice the size of the private key.

http://cryptoman.com/elliptic.htm has some basics.

mdxs:
Basically, this PR is just a retry of #34 (also titled "Support for SEC compressed keys"') by @samrushing in 2014/2015; based on code from https://github.com/richardkiss/pycoin and referring to the following for more information:

http://blog.richardkiss.com/?p=371
http://www.secg.org/
http://www.secg.org/sec1-v2.pdf

In addition to the `test_sec()` in `test_pyecdsa.py` from #34, I've also added uncompressed / compressed tests for to_sec/from_sec in the existing `test_pubkey_strings()` method.

Furthermore, I've made a little performance improvement in the `VerifyingKey.to_sec()` implementation and I've tried to ensure that the proposed changes are PEP8 compliant.

As noted in #53, this might address #53 needs and close #1 at the same time.

Please note that I'm fine if #34 is adopted first (credit to @samrushing where credit is due); after which I can adjust this PR for incremental improvements.

### DIFF
--- a/ecdsa/keys.py
+++ b/ecdsa/keys.py
@@ -1,13 +1,15 @@
 import binascii
 
 from . import ecdsa
+from . import ellipticcurve
 from . import der
 from . import rfc6979
 from .curves import NIST192p, find_curve
+from .numbertheory import square_root_mod_prime
+from .six import PY3, b
 from .util import string_to_number, number_to_string, randrange
 from .util import sigencode_string, sigdecode_string
 from .util import oid_ecPublicKey, encoded_oid_ecPublicKey
-from .six import PY3, b
 from hashlib import sha1
 
 
@@ -78,6 +80,32 @@ class VerifyingKey:
         assert point_str.startswith(b("\x00\x04"))
         return klass.from_string(point_str[2:], curve)
 
+    @classmethod
+    def from_sec(klass, string, curve=NIST192p, hashfunc=sha1,
+                 validate_point=True):
+        """Convert a public key in SEC binary format to a verifying key."""
+        # based on code from https://github.com/richardkiss/pycoin
+        if string.startswith(b('\x04')):
+            # uncompressed
+            return klass.from_string(string[1:], curve, hashfunc,
+                                     validate_point)
+        elif string.startswith(b('\x02')) or string.startswith(b('\x03')):
+            # compressed
+            is_even = string.startswith(b('\x02'))
+            x = string_to_number(string[1:])
+            order = curve.order
+            p = curve.curve.p()
+            alpha = (pow(x, 3, p) + (curve.curve.a() * x) + curve.curve.b()) % p
+            beta = square_root_mod_prime(alpha, p)
+            if is_even == bool(beta & 1):
+                y = p - beta
+            else:
+                y = beta
+            if validate_point:
+                assert ecdsa.point_is_valid(curve.generator, x, y)
+            point = ellipticcurve.Point(curve.curve, x, y, order)
+            return klass.from_public_point(point, curve, hashfunc)
+
     def to_string(self):
         # VerifyingKey.from_string(vk.to_string()) == vk as long as the
         # curves are the same: the curve itself is not included in the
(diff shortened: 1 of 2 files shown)


mdxs:
Quote from http://blog.richardkiss.com/?p=371 blog:

> The SEC ("Standards for Efficient Cryptography") format provides an alternate way of encoding the public key. This is the internal format that Bitcoin uses in transaction signatures to encode public keys. There is an uncompressed format, which has a prefix of a single 04 byte, followed by the x and y coordinates, and a compressed format, which has a prefix of 02 or 03 depending upon whether the y coordinate is even or odd, followed by the x coordinate.

trancee:
What's the holdup in merging this PR into the official release? I have been waiting for months for it to happen and was just wondering...

techguy613:
Happy New Year everyone! @warner, any idea when this might be merged? Having SEC-formatted keys would be very helpful in the library. Thanks for all your hard work on this module 😄

ecdsa/keys.py
```diff
@@ -78,6 +80,32 @@ def from_der(klass, string):
...
+            order = curve.order
+            p = curve.curve.p()
+            alpha = (pow(x, 3, p) + (curve.curve.a() * x) + curve.curve.b()) % p
+            beta = square_root_mod_prime(alpha, p)
+            if is_even == bool(beta & 1):
+                y = p - beta
+            else:
+                y = beta
+            if validate_point:
+                assert ecdsa.point_is_valid(curve.generator, x, y)
```
tomato42: assertions are [optimised out](https://docs.python.org/2/reference/simple_stmts.html#grammar-token-assert_stmt) when compiling

techguy613: In this case, what this means is that the klass.from_public_point() call will return a RuntimeError("Generator point has x or y out of range.") from ecdsa.Public_key initialization. What change would you make here?

mdxs: if I just add a "pass" it will work just fine, right?

mdxs: I made a small test using:
```
print "a"
if True:
    assert False
print "b"
```
and when I run that with -O it prints "a" and "b" showing that it "pass"es, while a non-optimised run will throw an AssertionError after printing "a" (as expected in this example).

So I actually think there is no need to change this PR, nor the current https://github.com/warner/python-ecdsa/blob/master/src/ecdsa/keys.py#L49

@tomato42 and @techguy613 can you confirm that there is no need to change the PR, or clarify why it should and how?

techguy613: I wouldn't change this PR as it is.

tomato42: it does need to to be changed - as asserts are optimised out, the call to `ecdsa.point_is_valid()` will be optimised out

now, either this call is necessary - then it's a big issue as the inputs may be attacker controlled - or this call is not necessary and then the whole `if` is superfluous

techguy613: Ok so why not remove the whole if statement then and the issue is resolved? The point is checked to be within the curve at a deeper level and will still throw an exception as I pointed out a little earlier.

mdxs: @tomato42: with your arguments that it needs to change, the current https://github.com/warner/python-ecdsa/blob/master/src/ecdsa/keys.py#L49 construct also needs a change. I will adjust this PR based on how that is (proposed to be) changed.

tomato42: @mdxs: that line 49 need to actually be changed to not use `assert` but the check needs to stay - the uncompressed point format can be malformed so checking if the point is on curve is needed

SimoneBronzini:
Any news on this PR? Is there any improvement left to be made before merging it?

florensie:
2 years later...

nomeata:
This would indeed be useful…

PR #60, opened 2016-10-27
Fix #59: Remove local six package

### LINKED ISSUE #59: Please do not bundle `six`
Is there any reason to bundle the `six` module? If it is only convenience for the user please remove it and add an `install_requires=['six']` to setup.py.

Bundling 3rd packages leads to duplicate code in a system and may lead to maintenance problems. And in your case not doing this saves you the work of updating the packages regularly :-)

sylvainpelissier:
Fix #59 : Remove local ecdsa.six package and update all files using the package.


warner:
Hm, having `setup.py` import `six` is a problem, because you have to manually install it before you can run e.g. `python setup.py install`.

We only use it in the speed tests, but there's no obvious way to ensure that it's been installed by that point. (and I've been cautioned against using `setup_requires=`, because it will cause problems down the road when pip wants to be more declarative).

I think the fix will be to change the way we run the tests (both unit tests and the speed benchmarks), by turning them into proper `tox` jobs, which can then run in a virtualenv where `six` has been installed. Then setup.py won't use six at all.

I'll do some more work on top of this branch, and see if I can move things over to `tox`.

warner:
Yup, got it. Pushing momentarily.

PR #71, opened 2017-06-27
Update Python versions

jonafato:
- Drop support for end-of-lifed Python 3.2 (see the [dev guide](https://docs.python.org/devguide/#status-of-python-branches) for end of life dates)
- Declare support and enable tests for Python 3.6


tomato42:
While 2.6 is not supported upstream, it is part of Red Hat Enterprise Linux 6, and thus it is supported.
And I'm using python-ecdsa there.

If there are problems getting it to work on 2.6, I'll gladly help. So until it's not possible to have 2.6 and current 3.x support, I'd really like for 2.6 support to remain.

jonafato:
My main goal here was adding 3.6 to the test suite. Happy to update this to put 2.6 back. 3.2 tends to have very little usage, and it was only listed in the Trove classifiers (no tests were running), so that one likely doesn't need to be there.

PR #78, opened 2018-01-23
add Python 3.6 support

tomato42:
make Python 3.6 support official


tomato42:
silly me, completely forgot about #71 

thanks for merging that other one!

PR #86, opened 2018-04-24
Specify markdown README for PyPI's long_description

dhimmel:
Set `long_description_content_type='text/markdown'` in `setup.py`. I believe this should trigger the README to display with proper formatting on [PyPI](https://pypi.org/project/ecdsa/). Currently, PyPI looks like:

pypi-ecdsa

Refs https://packaging.python.org/tutorials/distributing-packages/#description and https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,23 @@
 #!/usr/bin/env python
 
+import io
+import os
+
 from setuptools import setup
 import versioneer
 
 commands = versioneer.get_cmdclass().copy()
 
+# Use README.md to set markdown long_description
+directory = os.path.abspath(os.path.dirname(__file__))
+readme_path = os.path.join(directory, "README.md")
+with io.open(readme_path, encoding="utf-8") as read_file:
+    long_description = read_file.read()
+
 setup(name="ecdsa",
       version=versioneer.get_version(),
       description="ECDSA cryptographic signature library (pure python)",
+      long_description=long_description,
       long_description_content_type='text/markdown',
       author="Brian Warner",
       author_email="warner@lothar.com",
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ commands = versioneer.get_cmdclass().copy()
 setup(name="ecdsa",
       version=versioneer.get_version(),
       description="ECDSA cryptographic signature library (pure python)",
+      long_description_content_type='text/markdown',
       author="Brian Warner",
       author_email="warner@lothar.com",
       url="http://github.com/warner/python-ecdsa",


dhimmel:
@di is this pull request necessary for PyPI to render the README as markdown. Note that `setup.py` never sets `long_description`. I believe PyPI is recognizing the README from

https://github.com/warner/python-ecdsa/blob/ffe9a61390c76ff6f6636e29f26b45c900535aaf/MANIFEST.in#L2

If that's the case, will PyPI automatically use markdown formatting on the next release because of the `.md` extension?

di:
> If that's the case, will PyPI automatically use markdown formatting on the next release because of the .md extension?

@dhimmel No, you will need to set `long_description_content_type`. I would also recommend explicitly setting `long_description` as well, certain tools do not actually have the behavior you're describing.

(In fact, if this is happening for you now, I'd be curious what you're using to build/publish the distribution.)

Also, you're likely already aware, but you can use https://test.pypi.org/ to ensure that everything's set up correctly before actually making a new release on PyPI.

di:
Actually it seems like if you're using `flit` this might not actually be necessary, @takluyver would know for sure.

takluyver:
Yup, with flit you point to a file for the description ([like this](https://github.com/takluyver/flit/blob/0f0b9162d23f1d078cb3159503b5c1b78c438b03/pyproject.toml#L21)), and it uses the extension to select the content type appropriately (.rst, .md or .txt).

dhimmel:
> Yup, with flit you point to a file for the description, and it uses the extension to select the content type appropriately (.rst, .md or .txt).

Cool. So I'm not actually involved with this repository and am not sure if flit is being used. If it is, I'll close this PR. How do you find out?

takluyver:
It's not - if flit is in use, the repository will contain either a `pyproject.toml` (with the word flit in it) or `flit.ini`.

dhimmel:
>  I would also recommend explicitly setting long_description as well, certain tools do not actually have the behavior you're describing.

Let's wait for a project maintainer. Happy to make it so `setup.py` sets `long_description` to the text of `README.md` in this PR if that's requested.

dhimmel:
https://github.com/warner/python-ecdsa/pull/86/commits/0118ab4fb31a0cb5e08d9b68322bc33f88b91547 is the rebased commit. Will set `long_description` in next commit.

dhimmel:
> please update it so that long_description is set 

Done.

tomato42:
Perfect, thank you!

PR #87, opened 2018-05-30
fix the 'iff' typo

JoanJia:


tomato42:
r-

not a typo, "iff" stands for ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if)

PR #92, opened 2018-08-23
Add ability to include additional data as per RFC6979 section 3.6

junderw:
```
  Additional data may be added to the input of HMAC, concatenated
  after bits2octets(H(m)):
    K = HMAC_K(V || 0x00 || int2octets(x) || bits2octets(h1) || k')
```

Also follows implementation in secp256k1 library:
https://github.com/bitcoin-core/secp256k1/blob/452d8e4d2a2f9f1b5be6b02e18f1ba102e5ca0b4/src/secp256k1.c#L332

Recently, Bitcoin Core merged a feature that utilizes this feature in the secp256k1 library.

Eventually, Electrum (which depends on this library) may implement it, so would like to lay the ground work.

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -225,13 +225,13 @@ class SigningKey:
     def get_verifying_key(self):
         return self.verifying_key
 
-    def sign_deterministic(self, data, hashfunc=None, sigencode=sigencode_string):
+    def sign_deterministic(self, data, hashfunc=None, sigencode=sigencode_string, extra_entropy=b('')):
         hashfunc = hashfunc or self.default_hashfunc
         digest = hashfunc(data).digest()
 
-        return self.sign_digest_deterministic(digest, hashfunc=hashfunc, sigencode=sigencode)
+        return self.sign_digest_deterministic(digest, hashfunc=hashfunc, sigencode=sigencode, extra_entropy=extra_entropy)
 
-    def sign_digest_deterministic(self, digest, hashfunc=None, sigencode=sigencode_string):
+    def sign_digest_deterministic(self, digest, hashfunc=None, sigencode=sigencode_string, extra_entropy=b('')):
         """
         Calculates 'k' from data itself, removing the need for strong
         random generator and producing deterministic (reproducible) signatures.
@@ -239,7 +239,7 @@ class SigningKey:
         """
         secexp = self.privkey.secret_multiplier
         k = rfc6979.generate_k(
-            self.curve.generator.order(), secexp, hashfunc, digest)
+            self.curve.generator.order(), secexp, hashfunc, digest, extra_entropy=extra_entropy)
 
         return self.sign_digest(digest, sigencode=sigencode, k=k)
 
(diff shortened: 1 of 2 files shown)


src/ecdsa/keys.py
```diff
@@ -225,21 +225,21 @@ def to_der(self):
     def get_verifying_key(self):
         return self.verifying_key
 
-    def sign_deterministic(self, data, hashfunc=None, sigencode=sigencode_string):
+    def sign_deterministic(self, data, hashfunc=None, sigencode=sigencode_string, extra_entropy=b('')):
```
tomato42: the parentheses around apostrophes are unnecessary, `b''` is sufficient

src/ecdsa/keys.py
```diff
@@ -225,21 +225,21 @@ def to_der(self):
...
         hashfunc = hashfunc or self.default_hashfunc
         digest = hashfunc(data).digest()
 
-        return self.sign_digest_deterministic(digest, hashfunc=hashfunc, sigencode=sigencode)
+        return self.sign_digest_deterministic(digest, hashfunc=hashfunc, sigencode=sigencode, extra_entropy=extra_entropy)
 
-    def sign_digest_deterministic(self, digest, hashfunc=None, sigencode=sigencode_string):
+    def sign_digest_deterministic(self, digest, hashfunc=None, sigencode=sigencode_string, extra_entropy=b('')):
```
tomato42: same here, parentheses are unnecessary

src/ecdsa/rfc6979.py
```diff
@@ -56,7 +56,7 @@ def bits2octets(data, order):
 
 
 # https://tools.ietf.org/html/rfc6979#section-3.2
-def generate_k(order, secexp, hash_func, data):
+def generate_k(order, secexp, hash_func, data, extra_entropy=b('')):
```
tomato42: ditto

also, documentation missing for the new parameter

src/ecdsa/keys.py
junderw: I was following other portions of the code. (which use `from six import b` b())

If `b''` is fine by you. I will switch to it.

tomato42: ah, sorry, forgot about six, yes, for consistency it would need to be `b('')`, that being said, this six workaround it is necessary for compatibility with Python 2.5 and earlier, while the oldest officially supported by python-ecdsa is Python 2.6 which can handle the `b''` syntax fine

junderw:
Fixed the issues given.

Please let me know if you need anything else.

@warner @tomato42

junderw:
Thank you for your approval @tomato42 

I look forward to your input @warner

tomato42:
ok, so that's weird, but I don't see the travis run for this PR... (like, [at all](https://travis-ci.org/warner/python-ecdsa/pull_requests))

could you rebase on top of current master to see if that helps?

tomato42:
thanks for the patches!

SomberNight:
Would love to see a release that includes this.

tomato42:
@SomberNight list of items planned for [0.14](https://github.com/warner/python-ecdsa/milestone/1) is public, at the very least, until proper test coverage is done, 0.14 is not happening (I'd consider releasing 0.14 if only features are left in that milestone, but good test coverage is needed)

Issue #36, opened 2015-01-03
RFC6979 proper handling of bad r or s.


dabura667:
https://github.com/warner/python-ecdsa/blob/master/ecdsa/ecdsa.py#L147

Currently deterministic signing will bring the code to this point and fail in the (EXTREMELY) unlikely event of r == 0 or s == 0.

However, RFC 6979 states that:

> If that value of k is within the [1,q-1] range, and **is
> suitable for DSA or ECDSA (i.e., it results in an r value
> that is not 0**; see Section 3.4), then the generation of k is
> finished. The obtained value of k is used in DSA or ECDSA.
> **Otherwise, compute**:

So if the user is using `keys.SigningKey.sign_deterministic()` and `ecdsa.Private_key.sign()` returns an invalid r or s, the Step H loop should repeat once more.

Bitcore, for example, uses an int called badrs in the signing function
https://github.com/bitpay/bitcore/blob/master/lib/crypto/ecdsa.js#L204

In the event that r or s is 0, badrs is incremented, and they pass that int to the RFC6979 function to force a loop here:
https://github.com/bitpay/bitcore/blob/master/lib/crypto/ecdsa.js#L100

This is for efficiency, as adding a r and s value check would add an extra calculation of r and s.

However, I think it would be trivial to just add a function checking the r and s value in the current check here, and I don't think it is that much of a computational waste.
https://github.com/warner/python-ecdsa/blob/master/ecdsa/rfc6979.py#L99

Seeing as the `ecdsa.Private_key.sign()` does not necessarily use `keys.SigningKey.sign_deterministic()` it would be better to avoid a method like the `badrs` and create a new function for use in line 99 in rfc6979.py to check if r or s will be 0.

Maybe instead of only passing the curve order to generate_k(), you could pass the entire curve object, and pass along that object to the r and s check function. (which will undoubtedly need the generator, order, etc...)

It's a bit nit-picky, but still. It's written in the RFC.

Thanks in advanced for your thoughts.

I refrained from a pull request, as I first wanted to ask how you would plan to attack this problem.

dabura667:
https://github.com/dabura667/python-ecdsa/compare/warner:master...dabura667:master

I have implemented my vision of how it would work on my fork's master branch.

PR #93, opened 2018-08-30
Make sure that r and s don't become 0

junderw:
As per rfc6979:

> If that value of k is within the [1,q-1] range, **and is**
> **suitable for DSA or ECDSA (i.e., it results in an r value**
> **that is not 0; see Section 3.4)**, then the generation of k is
> finished. The obtained value of k is used in DSA or ECDSA.
> Otherwise, compute:

` k is within the [1,q-1] range` is being checked, but the part about r and s values aren't 0 is not currently being checked.

In the event a sig with 0 as the r or s value is leaked, the private key is easily calculated.

(Though this is extremely rare, this PR is for correctness sake)

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -1,9 +1,10 @@
-import binascii, re
+import binascii
 
 from . import ecdsa
 from . import der
 from . import rfc6979
 from .curves import NIST192p, find_curve
+from .ecdsa import RSZeroError
 from .util import string_to_number, number_to_string, randrange
 from .util import sigencode_string, sigdecode_string
 from .util import oid_ecPublicKey, encoded_oid_ecPublicKey
@@ -249,11 +250,8 @@ class SigningKey:
             try:
                 r, s, order = self.sign_digest(digest, sigencode=simple_r_s, k=k)
                 break
-            except RuntimeError as e:
-                if re.match("^amazingly unlucky random number (r|s)$", str(e)):
-                    retry_gen += 1
-                else:
-                    raise e
+            except RSZeroError:
+                retry_gen += 1
 
         return sigencode(r, s, order)
 
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -240,14 +240,14 @@ class SigningKey:
         secexp = self.privkey.secret_multiplier
 
         def simple_r_s(r, s, order):
-            return r, s, order
+            return r, s
 
         retry_gen = 0
         while True:
             k = rfc6979.generate_k(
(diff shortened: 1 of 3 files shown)


junderw:
Fixes #36

src/ecdsa/rfc6979.py
```diff
@@ -56,12 +56,13 @@ def bits2octets(data, order):
...
 
 # https://tools.ietf.org/html/rfc6979#section-3.2
-def generate_k(order, secexp, hash_func, data):
+def generate_k(order, secexp, hash_func, data, fail_loops=0):
     '''
         order - order of the DSA generator used in the signature
         secexp - secure exponent (private key) in numeric form
         hash_func - reference to the same hash function used for generating hash
         data - hash in binary form of the signing data
+        fail_loops - integer of the number of times to loop due to failure
```
tomato42: it's not really a failure on `generate_k`'s part, and I think this makes is rather confusing

it's more like "retry_gen - int - how many good 'k' values to skip before returning"

dabura667:
Thanks!

junderw:
Renamed. Thanks!

junderw:
Found another super-rare-will-probably-never-happen bug.

ecdsa.Private_key.sign was not performing `mod n` on the x value of k * G.

The specification of ECDSA requires that the x value is mod n to create r...

It is just such a rare occurrence in all curves, that no one has hit this bug.

junderw:
Looking into ecdsa.Private_key.sign it was raising RuntimeErrors for r and s being 0... so instead I catch them, match the message, and if the message matches, I ignore the RuntimeError, uptick the retry counter, and continue looping.

src/ecdsa/keys.py
```diff
@@ -240,16 +240,20 @@ def sign_digest_deterministic(self, digest, hashfunc=None, sigencode=sigencode_s
...
         while True:
             k = rfc6979.generate_k(
                 self.curve.generator.order(), secexp, hashfunc, digest, retry_gen=retry_gen)
-            r, s = self.sign_digest(digest, sigencode=simple_r_s, k=k)
-            if r != 0 and s != 0:
+            try:
+                r, s, order = self.sign_digest(digest, sigencode=simple_r_s, k=k)
                 break
-            retry_gen += 1
+            except RuntimeError as e:
```
tomato42: personally, not a fan of using native exceptions like that... Maybe change the raised exception to something ecdsa specific and make the one raised in case of zero r or s inherit also from RuntimeError?

junderw: Done.

tomato42:
Thanks for the patches!

PR #97, opened 2018-09-08
Update Python versions

hugovk:
Like https://github.com/warner/python-ecdsa/pull/71, but one year on :)

Rather than removing EOL 3.2 and adding new 3.6, this removes EOL 3.3 (since [2017-09-29](https://devguide.python.org/#status-of-python-branches)) and adds new 3.7.

Python 3.7 needs a little workaround on Travis CI to use Xenial with sudo, see 
https://github.com/travis-ci/travis-ci/issues/9815.

This also adds `python_requires` to setup.py to help pip.

I've left EOL Python 2.6 in place, is it still needed for RHEL 6?

---

By the way, here's the pip installs for ecdsa from PyPI for August 2018:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  70.70% |        687,544 |
| 3.6            |  17.94% |        174,475 |
| 3.5            |   7.17% |         69,774 |
| 3.4            |   2.18% |         21,182 |
| 3.7            |   1.42% |         13,849 |
| 2.6            |   0.56% |          5,403 |
| 3.3            |   0.02% |            195 |
| 3.2            |   0.00% |             44 |
| 3.8            |   0.00% |             16 |
| None           |   0.00% |              4 |
| Total          |         |        972,486 |

Source: `pypinfo --start-date 2018-08-01 --end-date 2018-08-31 -md --percent ecdsa pyversion`


tomato42:
I just fixed CI on 3.3 so am not so eager to drop it :smiley: (also, I'm using this package with tlsfuzzer and tlslite-ng, which still officially support 3.2 because of Debian wheezy)

In other words, unless there's a good reason to drop 3.3 (except its upstream support status), I am willing to do the extra work to keep CI and `python-ecdsa` working on 3.3. Given that 2.6 isn't going away, I don't think dropping 3.3 gives us access to any additional features (please correct me if I'm wrong, I haven't researched it).

adding 3.7 to the testing mix is much appreciated though

hugovk:
Yeah, there's nothing much special to be gained from dropping 3.3, except not needing to worry about pinning to outdated dependencies, and freeing a CI slot.

I've dropped the 3.3 dropping :)

tomato42:
@hugovk thanks!

tomato42:
BTW: Support for RHEL-6 ends in 2024, so yes, py2.6 support will be needed for quite some time still.

hugovk:
I see extended EOL of RHEL 5 is 2020, but I don't see many projects maintaining support for Python 2.4.

I wonder, is the adoption of RHEL 5 is much lower compared to RHEL 6, or something else?

Thanks!

tomato42:
for RHEL 5 there already is a pre-packaged Python 2.6 in EPEL: https://ftp-stud.hs-esslingen.de/pub/Mirrors/archive.fedoraproject.org/epel/5/x86_64/python26-2.6.8-2.el5.x86_64.rpm

PR #100, opened 2018-12-17
Add encoding for Unicode to prevent TypeError

hskang9:
TypeError: Unicode-objects must be encoded before hashing

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ signature string: it either returns True or raises BadSignatureError.
 from ecdsa import SigningKey
 sk = SigningKey.generate() # uses NIST192p
 vk = sk.get_verifying_key()
-signature = sk.sign("message".encode())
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,8 +102,8 @@ signature string: it either returns True or raises BadSignatureError.
 from ecdsa import SigningKey
 sk = SigningKey.generate() # uses NIST192p
 vk = sk.get_verifying_key()
-signature = sk.sign("message")
-assert vk.verify(signature, "message")
+signature = sk.sign("message".encode())
+assert vk.verify(signature, "message".encode())
```
tomato42: I think that changing it to byte string: `b"message"` would be less ambiguous what the intent is

tomato42:
Looks good, thank you!

Issue #108, opened 2019-07-17
`from_public_key_recovery` function using wrong parameter


kcorlidy:
pubkey = "fe27389aada80332ec2478814eb3c81de9d1109bc3c21d23853a43bd16ed176057371f3ce271d1ef423ef6fa56f97925a07036158bf90bb87c6c24f248932554"

	msg = "3045022100c3c38c977a3c11214f76eb59616a69838dbc14fc0d666e16c0d65989f7f9245b022037861ec5d4599caf17d80da102a89616a545cfca1932253bc2a539a0c9cb66e8"
	msg = bytes.fromhex(msg)
	raw_msg = "9302bda273a887cb40c13e02a50b4071a31fd3aae3ae04021b0b843dd61ad18e"
	raw_msg = bytes.fromhex(raw_msg)

	# right data
	_r = 0x00c3c38c977a3c11214f76eb59616a69838dbc14fc0d666e16c0d65989f7f9245b
	_s = 0x37861ec5d4599caf17d80da102a89616a545cfca1932253bc2a539a0c9cb66e8
	signature = msg
	data = raw_msg
	curve = ecdsa.curves.SECP256k1

	print("correct", _r, _s)
	generator = curve.generator
	r, s = ecdsa.util.sigdecode_der(signature, generator.order())
	print(r, s)

	pubkey = ecdsa.VerifyingKey.from_public_key_recovery(
			signature=signature, data=data, curve=curve, sigdecode=ecdsa.util.sigdecode_der)
	print(hexlify(pubkey[0].to_string()), hexlify(pubkey[1].to_string()))

Then i got error

	correct 88546510979682037774707173221481493851486333744139811890585270440885638210651 25114176564622922097833533925221099216470038098832442934481636541806701930216

	88546510979682037774707173221481493851486333744139811890585270440885638210651 25114176564622922097833533925221099216470038098832442934481636541806701930216
	Traceback (most recent call last):
	  File "C:\Users\hasee\Desktop\test2.py", line 85, in <module>
	    signature=signature, data=data, curve=curve, sigdecode=ecdsa.util.sigdecode_der)
	  File "D:\anaconda3\envs\py37\lib\site-packages\ecdsa\keys.py", line 89, in from_public_key_recovery
	    return klass.from_public_key_recovery_with_digest(signature, digest, curve, hashfunc=sha1, sigdecode=sigdecode_string)
[...]

PR #109, opened 2019-07-17
To fix issue #108

kcorlidy:
fixed the issue that [keys.py#Line89](https://github.com/warner/python-ecdsa/blob/master/src/ecdsa/keys.py#L89) using the wrong parameter

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -86,7 +86,7 @@ class VerifyingKey:
         # returns a list of verifying keys for this signature and message
         
         digest = hashfunc(data).digest()
-        return klass.from_public_key_recovery_with_digest(signature, digest, curve, hashfunc=sha1, sigdecode=sigdecode_string)
+        return klass.from_public_key_recovery_with_digest(signature, digest, curve, hashfunc=sha1, sigdecode=sigdecode)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


kcorlidy:
> looks good, but I'd rather not merge with CI failing
> 
> could you try fixing the CI in separate PR (I'll be busy for the next week, so the earliest I could try fixing it myself is next Friday)?

It looks like the python3.3 could not run the tox version which above 2.5(2.5 is fine), https://travis-ci.org/kcorlidy/python-ecdsa/builds/560858735.

kcorlidy:
It looks like the python3.3 could not run the tox version which above 2.5, because there is no `contextlib.suppress`. through my test, tox <= 2.5 failed, <2.5 succeed, so the latest one python3.3 can use might be tox 2.4.1. https://travis-ci.org/kcorlidy/python-ecdsa/builds/560858735.

kcorlidy:
I ever tried to install `contextlib2` first, but the result was unexpected it did not work, because `importlib-metadata` still chooses to use `contextlib`. Now, my **A choice** is downgrade `tox` to **2.4.1**. **B choice** is downgrade `pluggy` to **0.5.2**. 
 - In situation A, can only use tox 2.4.1.
 - In situation B, keep using tox 2.9.1(tox<3). So i suggest this one. [my test](https://travis-ci.org/kcorlidy/python-ecdsa/jobs/560913186)

tomato42:
I don't have a preference, either option is fine

PR #115, opened 2019-09-26
Robust signature decoding

### LINKED ISSUE #114: Inconsistent handling of malformed DER signatures
When the DER signature is malformed, the `VerifyingKey.verify(sigdecode=sigdecode_der)` raises unexpected exceptions and in some cases does not raise any exception.

`UnexpectedDER`, `IndexError` and `AssertionError` are raised in addition to `BadSignatureError`, at the same time flipping the 4th or 6th least significant bit of 2nd byte of a NIST521p signature does not cause it to be rejected (the DER encoding is strict, and every value has one defined encoding, any other encoding is invalid and needs to be rejected)

tomato42:
Ensure that the signature verification raises the same exception irrespective if the error is caused by mismatch of the hash value or because the encoding of the signature is invalid

fixes #114

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+# workaround for 3.7 not available in default configuration
+# travis-ci/travis-ci#9815
+dist: trusty
 sudo: false
 language: python
 cache: pip
--- a/build-requirements-2.6.txt
+++ b/build-requirements-2.6.txt
@@ -1,3 +1,4 @@
 tox
 coveralls<1.3.0
 idna<2.8
+unittest2
--- a/src/ecdsa/der.py
+++ b/src/ecdsa/der.py
@@ -75,10 +75,15 @@ def remove_constructed(string):
 
 
 def remove_sequence(string):
+    if not string:
+        raise UnexpectedDER("Empty string does not encode a sequence")
     if not string.startswith(b("\x30")):
-        n = string[0] if isinstance(string[0], integer_types) else ord(string[0])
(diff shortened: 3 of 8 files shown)


rjrelyea:
it basically looks good with one exception:
line 138 in function remove_integer. You check for nbytes < 0x80 and say 'Negative integers not supported'. I think this is checking that the length of the integer is less than 128 bytes (1024 bits), not that the first bit of the integer is on (indicating that the integer is negative).  This seems wrong. Either the error message or the code is wrong. I presume the latter since in RSA signatures, we often have integer lengths longer than 128 (like 2048 bit keys).

tomato42:
> You check for nbytes < 0x80 and say 'Negative integers not supported'. I think this is checking that the length of the integer is less than 128 bytes (1024 bits),

no, `nbytes` it's the first byte of the part of buffer that represents the integer – it's the most significant byte – I'll rename the variable to make it more readable

tomato42:
@rjrelyea: updated, I've also added more checks to verify if the values are correct DER encodings (not just BER), so please take a second look

rjrelyea:
OK, that looks good. nbytes was probably from the copy of dealing with the length field,
Your checks for minimal integer and minimal length encoding are correct.

opoplawski:
Is there an ETA for a 0.14 release?  I just got CVE bugs filed against this in Fedora.  Thanks.

tomato42:
@opoplawski actually, I'm working on 0.13.3 in #124 – there has been too little testing in master to consider it 0.14-worthy

Issue #95, opened 2018-09-05
Documentation fixes


tomato42:
A lot of documentation are comments, not documentation strings – they should be changed.

Most documentation strings don't actually document what the different arguments are and what the expected outputs are – using the sphinx format to document that would be nice.

Finally, many methods and functions are not documented at all, that should also be fixed.

tomato42:
`ecdsa` module should also redirect people to `keys`, as that's where high-level API lives

PR #117, opened 2019-09-28
Provide documentation for VerifyingKey and SigningKey

### LINKED ISSUE #56: Python 3: SigningKey.to_string() return a bytes object instead of a string
```
>>> import ecdsa
>>> type(ecdsa.SigningKey.generate().to_string())
<class 'bytes'>
```

Is this normal?

Example of how to get a string for someone that need to display a key:

```
>>> import codecs
>>> sk = ecdsa.SigningKey.generate().to_string()
>>> codecs.encode(sk, 'hex').decode("utf-8")  # maybe there is a simpler way
'5d9411d69b0286f51d5ef52300dcd02b617d843877c8a881'
```

tomato42:
Document the VerifyingKey, SigningKey and few directly related classes.

fixes #56 

work towards #95

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -596,23 +596,78 @@ class VerifyingKey(object):
 
 
 class SigningKey(object):
+    """
+    Class for handling keys that can create signatures (private keys).
+
+    :ivar ecdsa.curves.Curve curve: The Curve over which all the cryptographic
+        operations will take place
+    :ivar default_hashfunc: the function that will be used for hashing the
+        data. Should implement the same API as hashlib.sha1
+    :ivar int baselen: the length of a :term:`raw encoding` of private key
+    :ivar ecdsa.keys.VerifyingKey verifying_key: the public key
+        associated with this private key
+    :ivar ecdsa.ecdsa.Private_key privkey: the actual private key
+    """
+
     def __init__(self, _error__please_use_generate=None):
+        """Unsupported, please use one of the classmethods to initialise."""
         if not _error__please_use_generate:
             raise TypeError("Please use SigningKey.generate() to construct me")
+        self.curve = None
+        self.default_hashfunc = None
+        self.baselen = None
+        self.verifying_key = None
+        self.privkey = None
 
     @classmethod
-    def generate(klass, curve=NIST192p, entropy=None, hashfunc=sha1):
+    def generate(cls, curve=NIST192p, entropy=None, hashfunc=sha1):
+        """
+        Generate a random private key.
+
+        :param curve: The curve on which the point needs to reside, defaults
+            to NIST192p
+        :type curve: ecdsa.curves.Curve
+        :param entropy: Source of randomness for generating the private keys,
+            should provide cryptographically secure random numbers if the keys
+            need to be secure. Uses os.urandom() by default.
+        :type entropy: callable
+        :param hashfunc: The default hash function that will be used for
+            signing, needs to implement the same interface
+            as hashlib.sha1
+        :type hashfunc: callable
+
+        :return: Initialised SigningKey object
+        :rtype: SigningKey
+        """
         secexp = randrange(curve.order, entropy)
-        return klass.from_secret_exponent(secexp, curve, hashfunc)
-
-    # to create a signing key from a short (arbitrary-length) seed, convert
-    # that seed into an integer with something like
-    # secexp=util.randrange_from_seed__X(seed, curve.order), and then pass
-    # that integer into SigningKey.from_secret_exponent(secexp, curve)
+        return cls.from_secret_exponent(secexp, curve, hashfunc)
 
     @classmethod
-    def from_secret_exponent(klass, secexp, curve=NIST192p, hashfunc=sha1):
-        self = klass(_error__please_use_generate=True)
+    def from_secret_exponent(cls, secexp, curve=NIST192p, hashfunc=sha1):
+        """
+        Create a private key from a random integer.
+
+        Note: it's a low level method, it's recommended to use the
+        :func:`~SigningKey.generate` method to create private keys.
+
+        :param int secexp: secret multiplier (the actual private key in ECDSA).
+            Needs to be an integer between 1 and the curve order.
+        :param curve: The curve on which the point needs to reside
+        :type curve: ecdsa.curves.Curve
+        :param hashfunc: The default hash function that will be used for
+            signing, needs to implement the same interface
+            as hashlib.sha1
+        :type hashfunc: callable
+
+        :raises MalformedPointError: when the provided secexp is too large
+            or too small for the curve selected
+        :raises RuntimeError: if the generation of public key from private
+            key failed
+
+        :return: Initialised SigningKey object
+        :rtype: SigningKey
+        """
+        self = cls(_error__please_use_generate=True)
         self.curve = curve
         self.default_hashfunc = hashfunc
         self.baselen = curve.baselen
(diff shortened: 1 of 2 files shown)


tomato42:
@ansasaki please review

src/ecdsa/keys.py
```diff
@@ -222,21 +472,112 @@ def to_string(self, encoding="raw"):
...
+
+        The PEM header of the key will be `BEGIN PUBLIC KEY`.
+
+        The format of the key is described in the `from_der()` method.
+        This method supports only "named curve" encoding of keys.
+
+        :param str point_encoding: specification of the encoding format
+            of public keys. "uncompressed" is most portable, "compressed" is
+            smallest, "hybrid" is uncommon and unsupported by most
+            implementations, is as big as "uncompressed". "raw" will create
```
ansasaki: Why is "raw" allowed if it generates malformed files?

src/ecdsa/keys.py
```diff
@@ -222,21 +472,112 @@ def to_string(self, encoding="raw"):
...
 
-    def verify(self, signature, data, hashfunc=None, sigdecode=sigdecode_string):
+    def verify(self, signature, data, hashfunc=None,
+               sigdecode=sigdecode_string):
+        """
+        Verify a signature made over provided data.
+
+        Will hash `data` to verify the signature.
+
+        By default expects signature in raw encoding. Can also be used to
```
ansasaki: Should ':term:`raw encoding`' be used here?

src/ecdsa/keys.py
```diff
@@ -222,21 +472,112 @@ def to_string(self, encoding="raw"):
...
+        """
         hashfunc = hashfunc or self.default_hashfunc
         digest = hashfunc(data).digest()
         return self.verify_digest(signature, digest, sigdecode)
 
     def verify_digest(self, signature, digest, sigdecode=sigdecode_string):
+        """
+        Verify a signature made over provided hash value.
+
+        By default expects signature in raw encoding. Can also be used to
```
ansasaki: Should ':term:raw encoding' be used here?

src/ecdsa/util.py
```diff
@@ -212,13 +212,43 @@ def sigencode_strings(r, s, order):
...
 
 def sigencode_string(r, s, order):
+    """
+    Encode the signature to raw format.
+
+    :param int r: first parametr of the signature
+    :param int s: second parameter of the signature
+    :param int order: the order of curve over which the signature was computed
+
+    :return: DER encoding of ECDSA signature
```
ansasaki: It returns the raw encoding or DER encoding?

src/ecdsa/util.py
```diff
@@ -244,44 +274,113 @@ def sigencode_der_canonize(r, s, order):
...
     r = string_to_number_fixedlen(signature[:l], order)
     s = string_to_number_fixedlen(signature[l:], order)
     return r, s
 
 
 def sigdecode_strings(rs_strings, order):
+    """
+    Decode the signature from two strings.
+
+    First string needs to encode 'r', second needs to encode the
```
ansasaki: Encode how? In other parts the expected encoding was more specific.

src/ecdsa/keys.py
```diff
@@ -10,35 +70,80 @@
...
+        self.pubkey = None
 
     @classmethod
-    def from_public_point(klass, point, curve=NIST192p, hashfunc=sha1):
-        self = klass(_error__please_use_generate=True)
+    def from_public_point(cls, point, curve=NIST192p, hashfunc=sha1):
+        """
+        Initialise the object from a Point object.
+
+        This is a low-level method, generally you will not want to use it.
```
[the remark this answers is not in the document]
tomato42: both: programmer can decode the point himself or herself, but it's error-prone; generally the point will be encoded somehow, so the from_der, from_pem, from_string are more appropriate...

src/ecdsa/keys.py
tomato42: because there is no code to forbid it, filed #136 to fix that (I want to keep this PR docs and stylistic changes only)

tomato42:
thanks for review; will fix the issues I haven't commented on

tomato42:
@ansasaki updated, please re-check

src/ecdsa/util.py
```diff
@@ -212,13 +212,43 @@ def sigencode_strings(r, s, order):
 
 
 def sigencode_string(r, s, order):
+    """
+    Encode the signature to raw format (:term:`raw encoding`)
+
+    :param int r: first parametr of the signature
```
ansasaki: s/parametr/parameter/

src/ecdsa/util.py
```diff
@@ -212,13 +212,43 @@ def sigencode_strings(r, s, order):
...
+    :param int s: second parameter of the signature
+    :param int order: the order of curve over which the signature was computed
```
ansasaki: s/the order of curve/the order of the curve/

src/ecdsa/util.py
```diff
@@ -212,13 +212,43 @@ def sigencode_strings(r, s, order):
...
+
+        Ecdsa-Sig-Value ::= SEQUENCE {
+            r       INTEGER,
+            s       INTEGER
+        }
+
+    It's expected that this function will be used as a `sigencode=` parameter
+    in SigningKey.sign() method.
+
+    :param int r: first parametr of the signature
```
ansasaki: s/parametr/parameter

src/ecdsa/util.py
```diff
@@ -212,13 +212,43 @@ def sigencode_strings(r, s, order):
...
+    :param int s: second parameter of the signature
+    :param int order: the order of curve over which the signature was computed
```
ansasaki: s/the order of curve/the order of the curve/

src/ecdsa/keys.py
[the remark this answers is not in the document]
ansasaki: Thank you for the clarification.

tomato42:
> I found some typos.

@ansasaki good catch, can you verify?

tomato42:
(I've also "sphinxified" the cross-links and made them a bit more consistent)

src/ecdsa/keys.py
```diff
@@ -1,5 +1,65 @@
...
+
+    DER
+        Distinguished Encoding Rules, the way to encode :term:`ASN.1` objects
+        deterministically and uniquely into byte strings.
+
+    ASN.1
+        Abstract Syntax Notation 1 is a standard description language for
+        specifying serialisation and deserialisation of data structures in a
+        portable and cross-platform way.
+"""
```
ansasaki: Just a nit, the glossary could be alphabetically sorted.

src/ecdsa/keys.py
```diff
@@ -153,29 +337,79 @@ def from_der(klass, string):
         if empty != b(""):
             raise der.UnexpectedDER("trailing junk after pubkey pointstring: %s" %
                                     binascii.hexlify(empty))
-        # the point encoding is padded with a zero byte
-        # raw encoding of point is invalid in DER files
-        if not point_str.startswith(b("\x00")) or \
-                len(point_str[1:]) == curve.verifying_key_length:
+        # the bitsting encoding is padded with a zero byte
```
ansasaki: s/bitsting/bitstring/

src/ecdsa/keys.py
tomato42: it will be once it is processed by sphinx :)

tomato42:
Thank you!

PR #125, opened 2019-10-05
Apply nonce bit-length mitigation to stop timing leakage.

J08nY:
This should be a mitigation for the Minerva attack. See https://minerva.crocs.fi.muni.cz for more info.

### DIFF
--- a/src/ecdsa/ecdsa.py
+++ b/src/ecdsa/ecdsa.py
@@ -56,6 +56,7 @@ Written in 2005 by Peter Pearson and placed in the public domain.
 from six import int2byte, b
 from . import ellipticcurve
 from . import numbertheory
+from .util import bit_length
 
 
 class RSZeroError(RuntimeError):
@@ -171,7 +172,15 @@ class Private_key(object):
     G = self.public_key.generator
     n = G.order()
     k = random_k % n
-    p1 = k * G
+    # Fix the bit-length of the random nonce,
+    # so that it doesn't leak via timing.
+    # This does not change that ks = k mod n
+    ks = k + n
+    kt = ks + n
+    if bit_length(ks) == bit_length(n):
+      p1 = kt * G
+    else:
+      p1 = ks * G
     r = p1.x() % n
     if r == 0:
       raise RSZeroError("amazingly unlucky random number r")
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -86,6 +86,9 @@ class Point(object):
     else:
       return False
 
+  def __neg__(self):
+    return Point(self.__curve, self.__x, self.__curve.p() - self.__y)
+
   def __add__(self, other):
     """Add one point to another point."""
 
@@ -123,13 +126,12 @@ class Point(object):
       return result // 2
 
     e = other
-    if self.__order:
-      e = e % self.__order
-    if e == 0:
+    if e == 0 or (self.__order and e % self.__order == 0):
       return INFINITY
     if self == INFINITY:
       return INFINITY
-    assert e > 0
+    if e < 0:
+      return (-self) * (-e)
 
     # From X9.62 D.3.2:
 
--- a/src/ecdsa/curves.py
+++ b/src/ecdsa/curves.py
@@ -1,14 +1,18 @@
 from __future__ import division
 
 from . import der, ecdsa
+from .util import orderlen
+
+
+# orderlen was defined in this module previously, so keep it in __all__,
+# will need to mark it as deprecated later
+__all__ = ["UnknownCurveError", "orderlen", "Curve", "NIST192p",
+           "NIST224p", "NIST256p", "NIST384p", "NIST521p", "curves",
+           "find_curve"]
 
 
 class UnknownCurveError(Exception):
     pass
-
-
-def orderlen(order):
-    return (1+len("%x" % order))//2  # bytes
 
 
 # the NIST curves
--- a/src/ecdsa/rfc6979.py
+++ b/src/ecdsa/rfc6979.py
@@ -11,15 +11,13 @@ Many thanks to Coda Hale for his implementation in Go language:
 
 import hmac
 from binascii import hexlify
-from .util import number_to_string, number_to_string_crop
+from .util import number_to_string, number_to_string_crop, bit_length
 from six import b
 
 
-def bit_length(num):
-    # http://docs.python.org/dev/library/stdtypes.html#int.bit_length
-    s = bin(num)  # binary representation:  bin(-37) --> '-0b100101'
-    s = s.lstrip('-0b')  # remove leading zeros and minus sign
-    return len(s)  # len('100101') --> 6
+# bit_length was defined in this module previously so keep it for backwards
+# compatibility, will need to deprecate and remove it later
+__all__ = ["bit_length", "bits2int", "bits2octets", "generate_k"]
 
 
 def bits2int(data, qlen):
--- a/src/ecdsa/util.py
+++ b/src/ecdsa/util.py
@@ -5,7 +5,6 @@ import math
 import binascii
 from hashlib import sha256
 from . import der
-from .curves import orderlen
 from six import PY3, int2byte, b, next
 
 # RFC5480:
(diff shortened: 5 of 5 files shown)


tomato42:
Are you sure that this _fixes_ Minerva? it may decrease the side-channel, but I'm quite sure it doesn't eliminate it, see also: https://securitypitfalls.wordpress.com/2018/08/03/constant-time-compare-in-python/

src/ecdsa/ecdsa.py
```diff
@@ -171,7 +171,13 @@ def sign(self, hash, random_k):
     G = self.public_key.generator
     n = G.order()
     k = random_k % n
-    p1 = k * G
+    # Fix the bit-length of the random nonce,
+    # so that it doesn't leak via timing.
+    # This does not change that ks = k mod n
+    ks = k + n
+    if ks.bit_length() == n.bit_length():
```
tomato42: ints don't have `bit_length()` on python2.6, you can use `bit_length` from `rfc6979` module instead

src/ecdsa/ecdsa.py
```diff
@@ -171,7 +171,13 @@ def sign(self, hash, random_k):
...
+      ks = ks + n
```
tomato42: why it's added twice? it won't change its value `mod n`...

src/ecdsa/ellipticcurve.py
```diff
@@ -86,6 +86,9 @@ def __eq__(self, other):
     else:
       return False
 
+  def __neg__(self):
+    return Point(self.__curve, self.__x, self.__curve.p() - self.__y)
+
```
tomato42: why point negation is needed?

src/ecdsa/ecdsa.py
J08nY: That's exactly the point, `ks = k mod n`, yet the bit-length of `ks` is constant. See https://eprint.iacr.org/2011/232 and their mitigation in the Conclusion section.

J08nY:
> Are you sure that this _fixes_ Minerva? it may decrease the side-channel, but I'm quite sure it doesn't eliminate it, see also: [securitypitfalls.wordpress.com/2018/08/03/constant-time-compare-in-python](https://securitypitfalls.wordpress.com/2018/08/03/constant-time-compare-in-python/)

It is a compare of rather very small ints, both of which might be in the small int cache taht Python maintains, the leakage there is so small that the side-channel is thwarted by other noise. The example you sent is a string compare, which obviously is not constant time.

The conditional addition is what could be a problem, so I pushed a change that fixes that.

src/ecdsa/ellipticcurve.py
J08nY: Not really for the fix, but the scalarmult can handle negative scalars now.

tomato42:
My article is not just about comparison, _all_ integer operations, be it addition, multiplication or calculating powers in python are using variable bit-length arithmetic. I focused on small integers because that's what's needed for implementing Lucky 13 mitigations. But if small integers don't use a fast path, large integers won't use it either. That means that it is impossible to create constant-time implementation of any arithmetic in python. This is before even using anything but a naïve implementation of elliptic curve arithmetic.

> the leakage there is so small that the side-channel is thwarted by other noise

The https://eprint.iacr.org/2011/232.pdf article doesn't talk about python implementation, so I will need to see something more specific before considering it anything more than a scientific theory.

I am able to see in Lucky13 mitigation code written in python correlation of time with the value of padding length byte, code that hashes 64KiB of data using SHA-256 in the process of verifying the TLS CBC ciphertext. I'd consider that a lot of noise.

> The conditional addition is what could be a problem, so I pushed a change that fixes that.

this assumes the rest of the code is constant-time, as [security section](https://github.com/warner/python-ecdsa#security) in readme states, that is _not_ the case

See that multiplication by a scalar with the same bit length, but higher hamming weight requires more time:
```python
In [14]: from ecdsa.ecdsa import generator_256

In [15]: %timeit generator_256 * 0x8000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.8 ms ± 35.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [16]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.9 ms ± 58.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [17]: %timeit generator_256 * 0x8000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.3 ms ± 48.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [18]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.7 ms ± 16.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [19]: %timeit generator_256 * 0x8000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.4 ms ± 41.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [20]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.8 ms ± 19.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [21]: %timeit generator_256 * 0x8000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.9 ms ± 7.59 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [22]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.9 ms ± 3.81 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [23]: %timeit generator_256 * 0x8000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.6 ms ± 62.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [24]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000                                                                                                      
16.7 ms ± 46.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

J08nY:
Yes, the scalar mult here obviously leaks Hamming weight, but that is not exploitable via a HNP-like construction where you need to be sure of your guess of consecutive bits of many nonces.

This PR fixes an easily exploitable full private-key recovery vulnerability, I am not saying that it fixes all secret-dependent timing leaks, just this one. I don't see how moving towards not leaking so much easily exploitable information is controversial. An attacker would have a much harder time exploiting the remaining leaks than this one. (If you would get a HNP-like private key recovery attack from Hamming Weight of nonces, it would surely make for an interesting paper)

Sure, the README has a note saying that this is not constant time and might leak, but that is almost useless, I can install this package and start using it without ever seeing the README(and/or understanding why timing is an issue).

tomato42:
> Yes, the scalar mult here obviously leaks Hamming weight, but that is not exploitable via a HNP-like construction where you need to be sure of your guess of consecutive bits of many nonces.

hmm, I didn't think how strong the signal would be from the position of the second bit, and it doesn't look like it is very strong (I'm quite sure it's there, it's just harder to show it):

```python
In [34]: %timeit generator_256 * 0x9000000000000000000000000000000000000000000000000000000000000000
16.6 ms ± 38.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [35]: %timeit generator_256 * 0xc000000000000000000000000000000000000000000000000000000000000000
16.6 ms ± 38.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [36]: %timeit generator_256 * 0x8100000000000000000000000000000000000000000000000000000000000000
16.6 ms ± 32.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

> I don't see how moving towards not leaking so much easily exploitable information is controversial.

because it is futile for a library that has "being implemented in pure python" as the primary goal

but given how simple the fix is, I'm ok with merging it; but please fix the python 2.6 compatibility issue

> Sure, the README has a note saying that this is not constant time and might leak, but that is almost useless, I can install this package and start using it without ever seeing the README(and/or understanding why timing is an issue).

well, if anybody uses software for security applications without reading its documentation, I have bad news for them... that being said, any suggestion for the improvement of that text? or you think that addition of something like "if the attacker is able to measure the time it takes to generate the key or make signature, he or she will be able to recover the private key" or should we go straight for "This library, like all cryptographic libraries implemented in pure python, is insecure. Please use python-cryptography instead."?

PR #126, opened 2019-10-06
More test coverage

tomato42:
Uplift the test coverage from #124 to master
deprecate some unused methods
clean-up raised exceptions

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -48,12 +48,14 @@ class VerifyingKey:
     @staticmethod
     def _from_raw_encoding(string, curve, validate_point):
         order = curve.order
-        assert (len(string) == curve.verifying_key_length), \
-               (len(string), curve.verifying_key_length)
+        # real assert, from_string() should not call us with different length
+        assert len(string) == curve.verifying_key_length
         xs = string[:curve.baselen]
         ys = string[curve.baselen:]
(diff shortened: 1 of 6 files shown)


tomato42:
@ueno: please review

src/ecdsa/keys.py
```diff
@@ -48,12 +48,14 @@ def from_public_point(klass, point, curve=NIST192p, hashfunc=sha1):
...
-        assert (len(string) == curve.verifying_key_length), \
-               (len(string), curve.verifying_key_length)
+        # real assert, from_string() should not call us with different length
+        assert len(string) == curve.verifying_key_length
         xs = string[:curve.baselen]
         ys = string[curve.baselen:]
-        assert len(xs) == curve.baselen, (len(xs), curve.baselen)
-        assert len(ys) == curve.baselen, (len(ys), curve.baselen)
+        if len(xs) != curve.baselen:
+            raise MalformedPointError("Unexpected length of encoded x")
```
ueno: wouldn't it be a bit friendly that the actual length is embedded in the message?

tomato42: the problem is that we support 3 different lengths: for raw encoding, uncompressed X9.62 and compressed X9.62 so the message would be very confusing if we provided so much data

PR #138, opened 2019-10-15
Correct handling of BIT STRING in DER

### LINKED ISSUE #120: bug in der.encode_bitstring()
the `ecdsa.der.encode_bitstring()` is incorrect, the DER encoding of type BIT STRING is as follows:

`0x03` - the type
`<length>` - the length of the following encoding of the bitstring, including the <number of unused bits>; short form for encodings < 127 octets long and long form for longer
`<number of unused bits>` - the number of _least significant_ bits that need to be ignored
`<padded bit string>` - actual bit string, with padding, where padding bits must be zero

the current code does not include the &lt;number of unused bits> automatically or is able to handle its values different than zero (the only usages are in `ecdsa.keys`, so at least in theory  we can fix it without preserving the API... but then, there is no native bitstring type in python, so the &lt;number of unused bits> must be provided by the caller, we could use that to differentiate between old and new calling convention)

current code:
https://github.com/warner/python-ecdsa/blob/bcf6afe422a2fb142433e5579908e9d662e28f5e/src/ecdsa/der.py#L32-L33
https://github.com/warner/python-ecdsa/blob/bcf6afe422a2fb142433e5579908e9d662e28f5e/src/ecdsa/der.py#L201-L208

tomato42:
the first byte of the bit string in DER is the number of unused bits in the bit string, parse it properly and give expected error values if the parsing fails

fixes #120 

Todo:
 - [x] fix the loss of coverage for `keys.py`

### DIFF
--- a/src/ecdsa/der.py
+++ b/src/ecdsa/der.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 import binascii
 import base64
+import warnings
 from six import int2byte, b, integer_types, text_type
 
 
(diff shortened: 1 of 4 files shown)


tomato42:
@t8m please review

src/ecdsa/der.py
```diff
@@ -29,8 +30,61 @@ def encode_integer(r):
...
+        warnings.warn("Legacy call convention used, unused= needs to be "
+                      "specified",
+                      DeprecationWarning)
+    elif unused is not None:
+        if not 0 <= unused <= 7:
+            raise ValueError("unused must be integer between 0 and 7")
+        if unused and not s:
+            raise ValueError("unused is non-zero but s is empty")
+        encoded_unused = int2byte(unused)
+        len_extra = 1
```
t8m: Would it be good idea to either check or override the padding of the last byte here if unused is nonzero?

tomato42: checking is probably better, will do

t8m:
OK, still approved after the rebase and fixes.

tomato42:
@t8m Thank you!

PR #139, opened 2019-10-16
don't allow for creation of malformed DER files

### LINKED ISSUE #136: do not create files with malformed point encoding
to_der and to_pem in keys.py allow "raw" point encoding, we probably should not allow that...

tomato42:
when encoding in ASN.1 the public point has to use uncompressed,
compressed or hybrid representation, raw is not allowed
so disallow it in API

fixes #136


tomato42:
@t8m please review

PR #140, opened 2019-10-19
Readme updates

### LINKED ISSUE #67: OpenSSL signatures no longer compatible with ecdsa RAW format (README outdated)
Hi,

While trying to convert from ecdsa to python cryptography (which wraps OpenSSL) I discovered that the current iteration of OpenSSL (libssl 1.0.2g+) returns a DER formatted signature value instead of a raw pair of 32octect numbers. (FWIW: the signature appears to be a Sequence of NamedTypes containing Integers)

If, say, a JWT that has a signature from a direct OpenSSL wrapper that is unaware of this is attempted to be run through ecdsa, it'll fail due to the signature length check*. Folks who wish to use this library should check signature length != 64 and perform whatever transmogrification required to get the raw pair of key values that ecdsa requires.

Also: might want to update the examples in the "OpenSSL Compatibility" section of the README to reflect this. 

I'm going to try to follow up with other library owners in the chain to make sure that they're aware or at least comment about this problem lest others develop the same drinking habit I seem to have.

*really wish that some of these returned more meaningful errors than "AssertionError(71, 64)"

tomato42:
Few things changed since 2008... Update the README and speed.py to make them more current.

fixes #67


README.md
```diff
@@ -15,47 +15,65 @@ into other protocols.
...
-and 521 bits. The "short names" for these curves, as known by the OpenSSL
-tool (`openssl ecparam -list_curves`), are: prime192v1, secp224r1,
-prime256v1, secp384r1, and secp521r1. It also includes the 256-bit curve used
-by Bitcoin, whose short name is secp256k1. No other curves are included, but
-it would not be too hard to add more.
+popular NIST "Suite B" GF(p) (_prime field_) curves, with key lengths of 192,
+224, 256, 384, and 521 bits. The "short names" for these curves, as known by
+the OpenSSL tool (`openssl ecparam -list_curves`), are: `prime192v1`,
+`secp224r1`, `prime256v1`, `secp384r1`, and `secp521r1`. It also includes the
+256-bit curve used by Bitcoin, whose short name is `secp256k1`. No other curves
```
simo5: It also uses the 256-bit curve `secp256k1` used by Bitcoin.

README.md
```diff
@@ -15,47 +15,65 @@ into other protocols.
...
+are included, but it would not be too hard to add support for more curves
```
simo5: would not be -> is not

README.md
```diff
@@ -15,47 +15,65 @@ into other protocols.
...
 
 ## Dependencies
 
 This library uses only Python and the 'six' package. It is compatible with
 Python 2.6, 2.7 and 3.3+. It also supports execution on the alternative
-implementaions like pypy and pypy3
+implementations like pypy and pypy3
 
 To run the OpenSSL compatibility tests, the 'openssl' tool must be on your
-$PATH. This release has been tested successfully against both OpenSSL 0.9.8o
```
simo5: line above: on your -> in your

README.md
```diff
@@ -76,73 +94,103 @@ more for the wrapper. To run them all, do this:
...
+Do not allow attackers to measure how long it takes you to generate a keypair
+or sign a message. Do not allow attackers to run code on the same physical
+machine when keypair generation or signing is taking place (this includes
+virtual machines). Do not allow attackers to measure how much power your
+computer uses while generating the keypair or signing a message. Do not allow
+attackers to measure RF interference coming from your computer while generating
+a keypair or signing a message. Note: just loading the private key will cause
+keypair generation. Other operations or attack vectors may also be
+vulnerable to attacks. **For a sophisticated attacker observing just one
+operation with private key will be sufficient to completely
```
simo5: with a private key

README.md
```diff
@@ -76,73 +94,103 @@ more for the wrapper. To run them all, do this:
...
+signature = sk.sign(b"message")
+assert vk.verify(signature, b"message")
 ```
 
-The SigningKey can be serialized into several different formats: the shortest
+The `SigningKey` can be serialized into several different formats: the shortest
 is to call `s=sk.to_string()`, and then re-create it with
 `SigningKey.from_string(s, curve)` . This short form does not record the
-curve, so you must be sure to tell from_string() the same curve you used for
+curve, so you must be sure to tell `from_string()` the same curve you used for
```
simo5: tell -> pass to

tomato42:
> The RF/power consumption part is a bit dramatic but hey, it's true :)

I've seen how-to's that used this library on Raspberry PI, on such embedded platforms that is not an unlikely venue of attack...

PR #146, opened 2019-10-27
better test coverage for ecdsa module

tomato42:


src/ecdsa/test_numbertheory.py
```diff
@@ -116,6 +173,39 @@ def st_num_square_prime(draw):
...
 
 class TestNumbertheory(unittest.TestCase):
+    def test_gcd(self):
+        assert gcd(3 * 5 * 7, 3 * 5 * 11, 3 * 5 * 13) == 3 * 5
+        assert gcd([3 * 5 * 7, 3 * 5 * 11, 3 * 5 * 13]) == 3 * 5
+        assert gcd(3) == 3
+
+    @unittest.skipUnless(HC_PRESENT,
+                         "Hypothesis 2.0.0 can't be made tolerant of hard to "
+                         "meet requirements (like `is_prime()`)")
```
simo5: Not really sure I understand what this means, might be clarified, for people not too familiar with hypothesis

simo5: That is, what is the actual issue ?

tomato42: when you select random integers, few of them will be primes, so the test case generation takes a lot of time; hypothesis doesn't like it when it takes few seconds

tomato42:
> when all you are doing is testing if the math library works correctly

yes, the maths library that is part of this package... there is no `lcm()`, `gcd()` or `inverse_mod()` in standard python... or am I missing something?

src/ecdsa/test_numbertheory.py
simo5: Pleas mention that the issue is a time issue then.

simo5:
> > when all you are doing is testing if the math library works correctly
> 
> yes, the maths library that is part of this package... there is no `lcm()`, `gcd()` or `inverse_mod()` in standard python... or am I missing something?

There is math.gcd() but lcm() is indeed not there, although lcm() is so trivial to create once you have gcd(). I thought lcm() was also there but now that I look at it, it turns out I also implemented it recently for some test elsewhere :-D
So my comment I guess apply only to gcd() at this point.
It's not a big deal.

tomato42:
> There is  `math.gcd()`

Indeed, but not in older versions:
https://docs.python.org/3/library/math.html#math.gcd: 
> New in version 3.5.

probably we can use it to speed up newer versions, but the old code needs to remain

ok, thanks for the review!

PR #149, opened 2019-10-28
add ecdh class

merlokk:


src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,22 @@ def sign_number(self, number, entropy=None, k=None):
...
+        Allows two parties, each having an elliptic-curve public–private key 
+        pair, to establish a shared secret over an insecure channel
+                
+        :param remote_public_key: public key of remote party (VerifyingKey)
+        :return: shared secret
+        """""
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privkey.secret_multiplier
+
+        sharing_secret = ecdsa.int_to_string(result.x())
```
tomato42: `ecdsa.util.number_to_string()` with the second parameter being the generator point order will handle padding automatically

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,22 @@ def sign_number(self, number, entropy=None, k=None):
...
+        """
+        Elliptic-curve Diffie–Hellman (ECDH). A key agreement protocol.
+        Allows two parties, each having an elliptic-curve public–private key 
+        pair, to establish a shared secret over an insecure channel
+                
+        :param remote_public_key: public key of remote party (VerifyingKey)
+        :return: shared secret
+        """""
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privkey.secret_multiplier
```
tomato42: result needs to be checked if it's not INFINITY and raise exception if it is

see 2nd action in [Section 3.3.1 of SEC1 v2](http://www.secg.org/sec1-v2.pdf)

src/ecdsa/test_pyecdsa.py
```diff
@@ -686,6 +686,16 @@ def test_VerifyingKey_encode_decode(curve, encoding):
     assert vk.pubkey.point == from_enc.pubkey.point
 
 
+def test_ecdh():
+    for vcurve in curves:
+        priv1 = SigningKey.generate(curve=vcurve, hashfunc=sha256)
+        priv2 = SigningKey.generate(curve=vcurve, hashfunc=sha256)
```
tomato42: why `hashfunc`?

src/ecdsa/keys.py
tomato42: the code also needs to check if `remote_public_key` is on the same curve

src/ecdsa/test_pyecdsa.py
```diff
@@ -686,6 +686,16 @@ def test_VerifyingKey_encode_decode(curve, encoding):
...
+
+        secret1 = priv1.ecdh_get_shared_secret(priv2.get_verifying_key())
+        secret2 = priv2.ecdh_get_shared_secret(priv1.get_verifying_key())
+        assert secret1 == secret2
+
+
```
tomato42: interoperability test with openssl?

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,22 @@ def sign_number(self, number, entropy=None, k=None):
         assert 1 <= _k < order
         sig = self.privkey.sign(number, _k)
         return sig.r, sig.s
+
+    def ecdh_get_shared_secret(self, remote_public_key):
```
tomato42: Like I said, I'd rather not have ECDH as part of VerifyingKey or SigningKey, if you'd prefer to work on this in steps, then I think this should be a standalone method in a new module (`ecdh.py`) and operate on `ecdsa.ecdsa.Public_key` and `ecdsa.ecdsa.Private_key` objects or a method of the `Private_key` class (`derive` or something)

src/ecdsa/test_pyecdsa.py
```diff
@@ -686,6 +686,16 @@ def test_VerifyingKey_encode_decode(curve, encoding):
     assert vk.pubkey.point == from_enc.pubkey.point
 
 
+def test_ecdh():
+    for vcurve in curves:
```
tomato42: please use pytest parametrize, like here:
https://github.com/warner/python-ecdsa/blob/4c92d31dec71c5fe478744d57d332d5ceb255a33/src/ecdsa/test_malformed_sigs.py#L35

then if one assert fails, the whole test won't be aborted

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,22 @@ def sign_number(self, number, entropy=None, k=None):
...
+        """
+        Elliptic-curve Diffie–Hellman (ECDH). A key agreement protocol.
```
tomato42: this em-dash (`–`) is outside 7-bit ASCII, please add encoding definition to the file

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,22 @@ def sign_number(self, number, entropy=None, k=None):
...
+        :return: shared secret
+        """""
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privkey.secret_multiplier
+
+        sharing_secret = ecdsa.int_to_string(result.x())
+        deltalen = self.curve.baselen - len(sharing_secret)
+        if deltalen > 0:
+            sharing_secret = b"\x00" * deltalen + sharing_secret
+        return sharing_secret
```
tomato42: typo? not `shared_secret`?

src/ecdsa/test_pyecdsa.py
merlokk: deleted)

src/ecdsa/test_pyecdsa.py
merlokk: added NIST tests with constants

src/ecdsa/keys.py
merlokk: dash removed)

merlokk:
strange) `Coverage increased (+0.4%)`

src/ecdsa/keys.py
tomato42: my worry is API stability: SigningKey is part of the public API. If we release a version with this method included we can't remove it without breaking the public API

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,28 @@ def sign_number(self, number, entropy=None, k=None):
...
+        pair, to establish a shared secret over an insecure channel
+                
+        :param remote_public_key: public key of remote party (VerifyingKey)
+        :return: shared secret
+        """""
+
+        if not self.curve.curve.contains_point(
+                remote_public_key.pubkey.point.x(),
+                remote_public_key.pubkey.point.y()):
+            raise ValueError("Invalid curve in public key.")
```
tomato42: this does not check if the if the point is a valid public key on the curve, for this two checks need to be performed:
https://github.com/warner/python-ecdsa/blob/ca11bcb62ae576e0fd2140fe1fa89ab90e3c5ec1/src/ecdsa/ellipticcurve.py#L75-L78

but combined with this code:
https://github.com/warner/python-ecdsa/blob/ca11bcb62ae576e0fd2140fe1fa89ab90e3c5ec1/src/ecdsa/ecdsa.py#L117-L118

it's always performed for points in the `Public_key` object (and since it's an expensive check we really don't want to do it twice)

second, this does not check if the _curves_ match, the code should simply compare `self.curve.curve` with `remote.curve.curve`

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,28 @@ def sign_number(self, number, entropy=None, k=None):
...
+
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privkey.secret_multiplier
+        if result == ellipticcurve.INFINITY:
+            raise ValueError("Invalid shared secret.")
```
tomato42: a custom exception (both here and above) would be more robust, something like `InvalidKeyShareError`

src/ecdsa/keys.py
```diff
@@ -1133,3 +1133,28 @@ def sign_number(self, number, entropy=None, k=None):
...
+
+        shared_secret = number_to_string(result.x(),
+                                         0xff << ((self.curve.baselen - 1) * 8))
```
tomato42: won't using `self.curve.order` be much simpler?

src/ecdsa/test_ecdh.py
```diff
@@ -0,0 +1,156 @@
...
+    secret2 = priv2.ecdh_get_shared_secret(priv1.get_verifying_key())
+    assert secret1 == secret2
+
+
+def test_ecdh_wrong_public_key_curve():
+    priv1 = SigningKey.generate(curve=NIST192p)
+    priv2 = SigningKey.generate(curve=NIST256p)
+
+    try:
+        priv1.ecdh_get_shared_secret(priv2.get_verifying_key())
```
tomato42: I think the following would be more readable:
```python
with pytest.raises(ValueError):
    priv1.ecdh_get_shared_secret(priv2.get_verifying_key())
```

src/ecdsa/test_ecdh.py
```diff
@@ -0,0 +1,156 @@
...
+        pytest.param(
+            NIST521p,
+            "017eecc07ab4b329068fba65e56a1f8890aa935e57134ae0ffcce802735151f4eac6564f6ee9974c5e6887a1fefee5743ae2241bfeb95d5ce31ddcb6f9edb4d6fc47",
+            "00685a48e86c79f0f0875f7bc18d25eb5fc8c0b07e5da4f4370f3a9490340854334b1e1b87fa395464c60626124a4e70d0f785601d37c09870ebf176666877a2046d01ba52c56fc8776d9e8f5db4f0cc27636d0b741bbe05400697942e80b739884a83bde99e0f6716939e632bc8986fa18dccd443a348b6c3e522497955a4f3c302f676",
+            "005fc70477c3e63bc3954bd0df3ea0d1f41ee21746ed95fc5e1fdf90930d5e136672d72cc770742d1711c3c3a4c334a0ad9759436a4d3c5bf6e74b9578fac148c831",
+            id="NIST521p"
+        ),
+    ],
+)
+def test_ecdh_(curve,privatekey,pubkey,secret):
```
tomato42: extra `_`?

tomato42:
> strange) `Coverage increased (+0.4%)`

I've merged two PRs to master, please rebase

src/ecdsa/keys.py
```diff
@@ -1168,3 +1180,25 @@ def sign_number(self, number, entropy=None, k=None):
...
+        :param remote_public_key: public key of remote party (VerifyingKey)
+        :return: shared secret
+        """""
+
+        if self.curve.curve != remote_public_key.curve.curve:
+            raise InvalidPublicKeyCurveError
+
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privkey.secret_multiplier
+        if result == ellipticcurve.INFINITY:
```
tomato42: we can test if this exception is raised by setting the `secret_multiplier` to either the value of order or 0 in the test case

merlokk:
done

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,94 @@
...
+    def LoadPrivateKeyFromStr(self, private_key, kcurve):
+        self.privateKey = SigningKey.from_string(private_key, curve=kcurve)
+        if not (self.publicKey is None) and \
+                self.publicKey.curve != kcurve:
+            self.publicKey = None
+
+    def GetMyPublicKey(self):
+        return self.privateKey.get_verifying_key()
+
+    def LoadPublicKey(self, public_key):
```
tomato42: I don't like calling the two different keys the same... yes, they are public keys from the point of ECC, but that's bad for usability

Maybe it should be a "Received Key" (peer's public key) and "Key To Share" (out public key)?

src/ecdsa/test_ecdh.py
```diff
@@ -0,0 +1,194 @@
...
+# https://github.com/scogliani/ecc-test-vectors/blob/master/ecdh_kat/secp192r1.txt
+# https://github.com/scogliani/ecc-test-vectors/blob/master/ecdh_kat/secp256r1.txt
+# https://github.com/coruus/nist-testvectors/blob/master/csrc.nist.gov/groups/STM/cavp/documents/components/ecccdhtestvectors/KAS_ECC_CDH_PrimitiveTest.txt
+@pytest.mark.parametrize(
+    "curve,privatekey,pubkey,secret",
+    [
+        pytest.param(
+            NIST192p,
+            "f17d3fea367b74d340851ca4270dcb24c271f445bed9d527",
+            "42ea6dd9969dd2a61fea1aac7f8e98edcc896c6e55857cc0dfbe5d7c61fac88b11811bde328e8a0d12bf01a9d204b523",
```
tomato42: nit: excessive line length

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,94 @@
...
+            raise InvalidPublicKeyCurveError(
+                "Curves for public key and private key is not equal.")
+        self.publicKey = public_key
+
+    def LoadPublicKeyFromStr(self, public_key_str):
+        return self.LoadPublicKey(
+            VerifyingKey.from_string(public_key_str, self.privateKey.curve))
+
+    def GenerateSharedSecret(self):
+        return number_to_string(
```
tomato42: maybe we should make the `number_to_string` a method, just like the `sigencode` in `SigningKey.sign()`, so that uses that don't need padding can easily get the value in format they need?

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.privateKey.privkey.secret_multiplier
+        if result == INFINITY:
+            raise InvalidSharedSecretError(
+                "Invalid shared secret (INFINITY).")
+
+        return result.x()
+
+    def GeneratePrivateKey(self, kcurve):
```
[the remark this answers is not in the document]
merlokk: there are 2 cases:
1. private key needs to be generated each time
2. private key needs to be the same each time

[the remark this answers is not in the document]
tomato42: I'd say that if one needs a different key each time, they should create a new ECDH object each time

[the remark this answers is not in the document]
merlokk: so) if i want to generate shared secret and have not imported a private key - in this case it needs to generate a new private key?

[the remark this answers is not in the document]
merlokk: or let it be like now?

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,120 @@
...
+        return self.load_shared_public_key(
+            VerifyingKey.from_string(public_key_str, self.curve))
+
+    def load_shared_public_key_der(self, public_key_str):
+        return self.load_shared_public_key(VerifyingKey.from_der(public_key_str))
+
+    def load_shared_public_key_pem(self, public_key_str):
+        return self.load_shared_public_key(VerifyingKey.from_pem(public_key_str))
+
+    def generate_sharedsecret_str(self):
```
tomato42: string is a misnomer from python2 days, `number_to_string` returns `bytes`

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,120 @@
...
+from .util import number_to_string
+from .ellipticcurve import INFINITY
+from .keys import SigningKey, VerifyingKey
+
+
+__all__ = ["ECDH", "NoKeyError", "InvalidCurveError",
+           "InvalidSharedSecretError"]
+
+
+class NoKeyError(AssertionError):
```
tomato42: why `AssertionError`?

merlokk:
```
class NoKeyError(AttributeError):
class NoCurveError(AttributeError):
class InvalidCurveError(RuntimeError):
class InvalidSharedSecretError(RuntimeError):
```

ok?

src/ecdsa/ecdh.py
merlokk: maybe `plain`?

src/ecdsa/ecdh.py
tomato42: others are inheriting from specific exceptions to preserve API (the code previously was raising AssertionError), but given this is new code, it can get it right the first time over

src/ecdsa/ecdh.py
tomato42: I've never seen `plain`... is it used in some other crypto library in similar context?

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,123 @@
...
+        self.public_key = None
+        if private_key:
+            self.load_private_key(private_key)
+        if public_key:
+            self.load_received_public_key(public_key)
+
+    def _get_shared_secret(self, remote_public_key):
+        if self.private_key is None:
+            raise NoKeyError(
+                "Private key needs to be set to create shared secret")
```
tomato42: why it needs to be set and can't be generated?

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.public_key is None:
+            raise NoKeyError(
+                "Public key needs to be set to create shared secret")
+        if self.private_key.curve != remote_public_key.curve:
```
tomato42: if we don't trust the other methods, wouldn't it be more prudent to do a `if not (self.private_key.curve == self.cruve == remote_public_key.curve)`?

src/ecdsa/ecdh.py
merlokk: there are 2 cases: with key generation and key import. and this 2 cases can be easily messed up. and because of that i think it needs to make it explicit.

src/ecdsa/ecdh.py
merlokk: added bytes there

merlokk:
what inconsistent?

tomato42:
> what inconsistent?

I meant the method names

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,123 @@
...
+        # shared secret = PUBKEYtheirs * PRIVATEKEYours
+        result = remote_public_key.pubkey.point * self.private_key.privkey.secret_multiplier
+        if result == INFINITY:
+            raise InvalidSharedSecretError(
+                "Invalid shared secret (INFINITY).")
+
+        return result.x()
+
+    def set_curve(self, key_curve):
+        self.private_key = SigningKey.generate(curve=key_curve)
```
tomato42: this does something completely unexpected given the name...

src/ecdsa/__init__.py
```diff
@@ -14,7 +16,8 @@
            "test_pyecdsa", "util", "six"]
 
 _hush_pyflakes = [SigningKey, VerifyingKey, BadSignatureError, BadDigestError,
-                  MalformedPointError, UnexpectedDER,
+                  MalformedPointError, UnexpectedDER, InvalidCurveError,
+                  NoKeyError, InvalidSharedSecretError, ECDH,
```
tomato42: `NoCurveError` missing

src/ecdsa/ecdh.py
tomato42: ok

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,123 @@
...
+from .util import number_to_string
+from .ellipticcurve import INFINITY
+from .keys import SigningKey, VerifyingKey
+
+
+__all__ = ["ECDH", "NoKeyError", "InvalidCurveError",
+           "InvalidSharedSecretError"]
+
+
+class NoKeyError(AttributeError):
```
tomato42: it's still using non-basic Exception (same for other exceptions here)

tomato42:
could you also rebase again on top of master? that should help with the failure in python 2.6 test coverage

tomato42:
please ping me when you address https://github.com/warner/python-ecdsa/pull/149#pullrequestreview-314508637

regarding documentation, I suggest reading [PEP 257](https://www.python.org/dev/peps/pep-0257/)

merlokk:
@tomato42 look please on method names. now it should more like `keys.py`

src/ecdsa/ecdh.py
```diff
@@ -54,9 +54,9 @@ def __init__(self, curve=None, private_key=None, public_key=None):
         self.private_key = None
         self.public_key = None
         if private_key:
-            self.load_private_key(private_key)
+            self.local_pk_set(private_key)
         if public_key:
-            self.load_received_public_key(public_key)
+            self.remote_vk_set(public_key)
```
tomato42: why did you change them?

merlokk: method names more consistent with `keys.py`

tomato42: I was talking about the `_str` vs `_bytes`; changing it to `_bytes` was what was ok with, so I didn't reiterate it in https://github.com/warner/python-ecdsa/pull/149#pullrequestreview-314508637

merlokk: ok. ill revert changes and start to do docstrings. ok?

tomato42: ok

merlokk:
interoperability with openssl:
```
zoom@mlinux:~/testecdh$ openssl pkeyutl -derive -inkey key.pem -peerkey pubkey.pem | xxd -p
8f457e34982478d1c34b9cd2d0c15911b72dd60d869e2cea
```

tomato42:
> interoperability with openssl:
> 
> ```
> zoom@mlinux:~/testecdh$ openssl pkeyutl -derive -inkey key.pem -peerkey pubkey.pem | xxd -p
> 8f457e34982478d1c34b9cd2d0c15911b72dd60d869e2cea
> ```
> 
> https://github.com/warner/python-ecdsa/pull/149/files#diff-165f92ad7e143f620eca07d4f8ffbe41R236

I meant something like this:
https://github.com/warner/python-ecdsa/blob/c3136e4250e7e2c53191acccbff8a936479536af/src/ecdsa/test_pyecdsa.py#L819-L852

merlokk:
done

src/ecdsa/ecdh.py
```diff
@@ -79,16 +79,27 @@ def _get_shared_secret(self, remote_public_key):
...
-        Set working curve for ecdh operation
+        Set working curve for ecdh operation.
         :param key_curve: curve from `curves`
         :return: none
         """
         self.curve = key_curve
 
     def generate_private_key(self):
+        """
+        Generate local private key for ecdh operation with curve that was set.
```
tomato42: what will happen if the curve was not set?

src/ecdsa/ecdh.py
```diff
@@ -79,16 +79,27 @@ def _get_shared_secret(self, remote_public_key):
...
+        :return: public (verifying) key from this private key
```
tomato42: type of returned object missing

src/ecdsa/ecdh.py
```diff
@@ -79,16 +79,27 @@ def _get_shared_secret(self, remote_public_key):
...
+        """
         return self.load_private_key(SigningKey.generate(curve=self.curve))
 
     def load_private_key(self, private_key):
+        """
+        Load private key from SigningKey (keys.py) class.
+        Needs to have the same curve as was set with set_curve method.
+        If curve is not set - it sets from this SigningKey
+        :param private_key: Initialised SigningKey class
```
tomato42: `:type private_key:` missing

src/ecdsa/ecdh.py
```diff
@@ -79,16 +79,27 @@ def _get_shared_secret(self, remote_public_key):
...
+        :return: public (verifying) key from this private key
```
tomato42: type of returned object missing
what exceptions it raises is missing

src/ecdsa/ecdh.py
```diff
@@ -79,16 +79,27 @@ def _get_shared_secret(self, remote_public_key):
...
     def generate_private_key(self):
+        """
+        Generate local private key for ecdh operation with curve that was set.
+        :return: public (verifying) key from this private key
+        """
         return self.load_private_key(SigningKey.generate(curve=self.curve))
 
     def load_private_key(self, private_key):
+        """
+        Load private key from SigningKey (keys.py) class.
```
tomato42: `Loads private key from SigningKey object.`?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
         return self.private_key.get_verifying_key()
 
     def load_private_key_bytes(self, private_key):
+        """
+        Load private key from plain bytes string.
```
tomato42: not only, `SingingKey.from_string()` will accept any bytes-like object, not only `bytes()`

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        Uses current curve and checks if key length corresponds to
+        the current curve.
+        :param private_key: private key in bytes string format
```
tomato42: `:type private_key:` missing
`:raises` missing

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
         return self.private_key.get_verifying_key()
 
     def load_private_key_bytes(self, private_key):
+        """
+        Load private key from plain bytes string.
+        Uses current curve and checks if key length corresponds to
+        the current curve.
```
tomato42: `Checks if the provided key matches the curve of ECDH key agreement.`?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        :param private_key: private key in bytes string format
+        :return: public (verifying) key from this private key
+        """
         return self.load_private_key(
             SigningKey.from_string(private_key, curve=self.curve))
 
     def load_private_key_der(self, private_key):
+        """
+        Load private key from DER string.
```
tomato42: s/string/byte string/ – on python3 it's an important distinction

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        Curve is get from DER and check if it the same as current curve
```
tomato42: `Compares the curve of the DER-encoded key with the ECDH set curve, uses the former if unset`?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+
+        Note, the only DER format supported is the RFC5915
+        Look at keys.py:SigningKey.from_der()
+
+        :param private_key: binary string with the DER encoding of private ECDSA key
+        :return: public (verifying) key from this private key
```
tomato42: again, type and raised exceptions missing, same for other below

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        """
         return self.load_private_key(SigningKey.from_der(private_key))
 
     def load_private_key_pem(self, private_key):
+        """
+        Load private key from PEM string.
+        Curve is get from PEM and check if it the same as current curve
```
tomato42: `Compares the curve of the DER-encoded key with the ECDH set curve, uses the former if unset`?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        it needs to have `EC PRIVATE KEY` section
+
+        :param private_key: text with PEM-encoded private ECDSA key
+        :return: public (verifying) key from this private key
+        """
         return self.load_private_key(SigningKey.from_pem(private_key))
 
     def get_public_key(self):
+        """
+        Get public key from local private key.
```
tomato42: `Provides a public key that matches the local private key`?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        Usually needs to send it to remote party.
```
tomato42: "usually"?

also, I think it should be "needs to be sent to the remote party"?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        :return: public (verifying) key from local private key
+        """
         return self.private_key.get_verifying_key()
 
     def load_received_public_key(self, public_key):
+        """
+        Load public key from VerifyingKey (keys.py) class.
```
tomato42: "object" rather than "class"? we need an instance after all...

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        the current curve.
+        :param public_key_str: public key in bytes string format
+        :return: none
+        """
         return self.load_received_public_key(
             VerifyingKey.from_string(public_key_str, self.curve))
 
     def load_received_public_key_der(self, public_key_str):
+        """
+        Load public key from DER binary string.
```
tomato42: "bytes" instead of "binary" – it's more common usage in python

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        Curve is get from DER and check if it the same as current curve
```
tomato42: again, "compared..."

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+
+        :param public_key_str: binary string with the DER encoding of public ECDSA key
+        :return: none
+        """
         return self.load_received_public_key(VerifyingKey.from_der(public_key_str))
 
     def load_received_public_key_pem(self, public_key_str):
+        """
+        Load public key from PEM string.
+        Curve is get from PEM and check if it the same as current curve
```
tomato42: same here

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        :param public_key_str: text with PEM-encoded public ECDSA key
+        :return: none
+        """
         return self.load_received_public_key(VerifyingKey.from_pem(public_key_str))
 
     def generate_sharedsecret_bytes(self):
+        """
+        Generate shared secret from local private key and remote public key.
+        Needs to have initialized private key and received public key before
+        generating.
```
tomato42: "The objects needs to have both private key and received public key before generation is allowed"?

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+        """
         return number_to_string(
                     self.generate_sharedsecret(),
                     self.private_key.curve.order)
 
     def generate_sharedsecret(self):
+        """
+        Generate shared secret from local private key and remote public key.
+        Needs to have initialized private key and received public key before
+        generating.
```
tomato42: same as above

src/ecdsa/ecdh.py
```diff
@@ -97,39 +108,123 @@ def load_private_key(self, private_key):
...
+
+        It the same for local and remote parties.
+        shared secret(local private key, remote public key ) ==
+                shared secret (local public key, remote private key)
+
+        :return: shared secret in int format
```
tomato42: `:rtype: int` would be cleaner than "in int format"

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,230 @@
...
+class InvalidSharedSecretError(Exception):
+    """ECDH. Raised in case the shared secret we obtained is an INFINITY."""
+
+    pass
+
+
+class ECDH(object):
+    """
+    Elliptic-curve Diffie-Hellman (ECDH). A key agreement protocol.
+    Allows two parties, each having an elliptic-curve public-private key
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    Elliptic-curve Diffie-Hellman (ECDH). A key agreement protocol.
+    Allows two parties, each having an elliptic-curve public-private key 
+    pair, to establish a shared secret over an insecure channel
+    """""
+
+    def __init__(self, curve=None, private_key=None, public_key=None):
+        """
+        ECDH init. Class can be init without parameters and then
+        it will needs to set all of them for proper operation.
```
tomato42: too long summary line

src/ecdsa/ecdh.py
```diff
@@ -0,0 +1,230 @@
...
+        :param curve: curve for operations
+        :param private_key: `my` private key for ecdh
+        :param public_key:  `their` public key for ecdh
```
tomato42: shouldn't `ecdh` be capitalised here?

merlokk:
done

src/ecdsa/test_ecdh.py
```diff
@@ -322,10 +319,18 @@ def test_ecdh_with_openssl(vcurve):
...
-        ssl_secret1 = e.read()
-    with open("t/secret1", "rb") as e:
-        ssl_secret2 = e.read()
+    try:
+        with open("t/secret1", "rb") as e:
+            ssl_secret1 = e.read()
+        with open("t/secret1", "rb") as e:
+            ssl_secret2 = e.read()
+    except FileNotFoundError:
+        pytest.skip("system openssl does not support `pkeyutl -derive`")
```
tomato42: I'm pretty sure that if openssl doesn't support -derive we won't get this far: the run_openssl will raise subprocess.SubprocessError exception because the exit code of command will be non-zero, won't it?

looking for `-derive` in `pkeyutl -help` output may be a better idea...

src/ecdsa/ecdh.py
```diff
@@ -38,18 +38,25 @@ class InvalidSharedSecretError(Exception):
...
     pair, to establish a shared secret over an insecure channel
     """""
 
     def __init__(self, curve=None, private_key=None, public_key=None):
         """
-        ECDH init. Class can be init without parameters and then
-        it will needs to set all of them for proper operation.
+        ECDH init.
+
+        Class can be init without parameters and then it will needs to set
```
tomato42: "Call can be initialised without parameters, then the first operation (loading either the key) will set the used curve.
All parameters must be ultimately set before share secret calculation will be allowed." ?

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
     def set_curve(self, key_curve):
         """
         Set working curve for ecdh operation.
```
tomato42: "Set the working curve for ecdh operations." ?

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
...
+
         :param key_curve: curve from `curves`
```
tomato42: I think "module" is missing here at the end

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
...
 
     def generate_private_key(self):
         """
         Generate local private key for ecdh operation with curve that was set.
-        :return: public (verifying) key from this private key
+
+        :raises NoCurveError: Curve must be set before key generation.
+
+        :return: public (verifying) key from this private key.
+        :rtype: byte string
```
tomato42: `self.private_key.get_verifying_key()` returns a byte string?

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
...
         """
+        if self.curve is None:
```
tomato42: the idiom usually used, when differentiation between None and empty collection is not needed, is to use `if not self.curve:`

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
...
+            raise NoCurveError("Curve must be set prior to key generation.")
         return self.load_private_key(SigningKey.generate(curve=self.curve))
 
     def load_private_key(self, private_key):
         """
-        Load private key from SigningKey (keys.py) class.
+        Load private key from SigningKey (keys.py) object.
         Needs to have the same curve as was set with set_curve method.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
+        if self.curve is None:
+            raise NoCurveError("Curve must be set prior to key load.")
         return self.load_private_key(
             SigningKey.from_string(private_key, curve=self.curve))
 
     def load_private_key_der(self, private_key):
         """
-        Load private key from DER string.
-        Curve is get from DER and check if it the same as current curve
+        Load private key from DER byte string.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
 
     def load_private_key_bytes(self, private_key):
         """
-        Load private key from plain bytes string.
-        Uses current curve and checks if key length corresponds to
-        the current curve.
+        Load private key from byte string.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
-        :return: public (verifying) key from this private key
+        :type private_key:   string
+        :return: public (verifying) key from this private key.
+        :rtype: byte string
         """
         return self.load_private_key(SigningKey.from_der(private_key))
 
     def load_private_key_pem(self, private_key):
         """
         Load private key from PEM string.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
         """
         return self.load_private_key(SigningKey.from_pem(private_key))
 
     def get_public_key(self):
         """
-        Get public key from local private key.
-        Usually needs to send it to remote party.
-        :return: public (verifying) key from local private key
-        """
+        Provides a public key that matches the local private key.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
+
+        :return: public (verifying) key from local private key.
+        :rtype: byte string
+       """
         return self.private_key.get_verifying_key()
 
     def load_received_public_key(self, public_key):
         """
-        Load public key from VerifyingKey (keys.py) class.
+        Load public key from VerifyingKey (keys.py) object.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
 
     def load_received_public_key_bytes(self, public_key_str):
         """
-        Load public key from bytes string.
+        Load public key from byte string.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
         :return: none
         """
         return self.load_received_public_key(
             VerifyingKey.from_string(public_key_str, self.curve))
 
     def load_received_public_key_der(self, public_key_str):
         """
-        Load public key from DER binary string.
-        Curve is get from DER and check if it the same as current curve
+        Load public key from DER byte string.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
+        :type public_key_str: byte string
         :return: none
         """
         return self.load_received_public_key(VerifyingKey.from_der(public_key_str))
 
     def load_received_public_key_pem(self, public_key_str):
         """
         Load public key from PEM string.
-        Curve is get from PEM and check if it the same as current curve
+        Compares the curve of the PEM-encoded key with the ECDH set curve,
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
         """
         return self.load_received_public_key(VerifyingKey.from_pem(public_key_str))
 
     def generate_sharedsecret_bytes(self):
         """
         Generate shared secret from local private key and remote public key.
-        Needs to have initialized private key and received public key before
-        generating.
-        :return: shared secret in binary string format
+        The objects needs to have both private key and received public key
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -218,13 +271,17 @@ def generate_sharedsecret_bytes(self):
     def generate_sharedsecret(self):
         """
         Generate shared secret from local private key and remote public key.
```
tomato42: empty line missing after summary line

src/ecdsa/ecdh.py
```diff
@@ -80,25 +87,39 @@ def _get_shared_secret(self, remote_public_key):
...
         If curve is not set - it sets from this SigningKey
+
         :param private_key: Initialised SigningKey class
-        :return: public (verifying) key from this private key
+        :type private_key: SigningKey
+
+        :raises InvalidCurveError: private_key curve not the same as self.curve
+
+        :return: public (verifying) key from this private key.
+        :rtype: byte string
```
tomato42: `self.private_key.get_verifying_key()` returns a byte string?

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
-        Load private key from plain bytes string.
-        Uses current curve and checks if key length corresponds to
-        the current curve.
+        Load private key from byte string.
+        Uses current curve and checks if the provided key matches
+        the curve of ECDH key agreement.
+
         :param private_key: private key in bytes string format
-        :return: public (verifying) key from this private key
+        :type private_key: byte string
```
tomato42:
```
:type private_key: :term:`bytes-like object`
```

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
+
+        :raises NoCurveError: Curve must be set before loading.
+
+        :return: public (verifying) key from this private key.
+        :rtype: byte string
```
tomato42: again, I don't think this is correct

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
+        Load private key from DER byte string.
+        Compares the curve of the DER-encoded key with the ECDH set curve,
+        uses the former if unset.
 
         Note, the only DER format supported is the RFC5915
         Look at keys.py:SigningKey.from_der()
 
         :param private_key: binary string with the DER encoding of private ECDSA key
-        :return: public (verifying) key from this private key
+        :type private_key:   string
```
tomato42:
```
:type private_key: :term:`bytes-like object`
```

src/ecdsa/ecdh.py
tomato42: also, I think referencing SigningKey method, so that the format of the byte string is defined, would be a good idea

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
 
         Note, the only PEM format supported is the RFC5915
         Look at keys.py:SigningKey.from_pem()
         it needs to have `EC PRIVATE KEY` section
 
         :param private_key: text with PEM-encoded private ECDSA key
-        :return: public (verifying) key from this private key
+        :type private_key: string
+        :return: public (verifying) key from this private key.
+        :rtype: byte string
```
tomato42: I don't think this is correct

src/ecdsa/ecdh.py
```diff
@@ -109,57 +130,78 @@ def load_private_key(self, private_key):
...
         """
-        Get public key from local private key.
-        Usually needs to send it to remote party.
-        :return: public (verifying) key from local private key
-        """
+        Provides a public key that matches the local private key.
+        Needs to be sent to the remote party.
+
+        :return: public (verifying) key from local private key.
+        :rtype: byte string
```
tomato42: again, don't think it's a byte string

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
-        Load public key from bytes string.
+        Load public key from byte string.
         Uses current curve and checks if key length corresponds to
         the current curve.
+
         :param public_key_str: public key in bytes string format
+        :type public_key_str: byte string
         :return: none
         """
         return self.load_received_public_key(
```
tomato42: if `load_received_public_key` returns None (doesn't return anything), why do we explicitly return it?

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
 
         Note, the only DER format supported is the RFC5912
         Look at keys.py:VerifyingKey.from_der()
 
-        :param public_key_str: binary string with the DER encoding of public ECDSA key
+        :param public_key_str: byte string with the DER encoding of public ECDSA key
+        :type public_key_str: byte string
         :return: none
         """
         return self.load_received_public_key(VerifyingKey.from_der(public_key_str))
```
tomato42: same here, why the `return`?

src/ecdsa/ecdh.py
```diff
@@ -169,47 +211,58 @@ def load_received_public_key(self, public_key):
...
+        Load public key from DER byte string.
+        Compares the curve of the DER-encoded key with the ECDH set curve,
+        uses the former if unset.
 
         Note, the only DER format supported is the RFC5912
         Look at keys.py:VerifyingKey.from_der()
 
-        :param public_key_str: binary string with the DER encoding of public ECDSA key
+        :param public_key_str: byte string with the DER encoding of public ECDSA key
+        :type public_key_str: byte string
```
tomato42: `public_key_str` names of parameters are part of public API, so I think dropping the `_str` part would make it cleaner

src/ecdsa/ecdh.py
```diff
@@ -218,13 +271,17 @@ def generate_sharedsecret_bytes(self):
     def generate_sharedsecret(self):
         """
         Generate shared secret from local private key and remote public key.
-        Needs to have initialized private key and received public key before
-        generating.
+        The objects needs to have both private key and received public key
+        before generation is allowed.
 
         It the same for local and remote parties.
```
tomato42: `It's the...` ?
also, I think it should be `party`, singular

README.md
```diff
@@ -466,3 +466,18 @@ comp_str = '022799c0d0ee09772fdd337d4f28dc155581951d07082fb19a38aa396b67e77759'
...
+from ecdsa import ECDH, NIST256p
+
+ecdh = ECDH(curve=NIST256p)
+ecdh.generate_private_key()
+local_public_key = ecdh.get_public_key()
+#send `local_public_key` to remote party and receive `remote_public_key` from remote party
+with open("remote_public_key.pem") as e:
+    remote_public_key = e.read()
+ecdh.load_received_public_key_pem(remote_public_key)
+secret = ecdh1.generate_sharedsecret_bytes()
```
tomato42: extra `1`?

src/ecdsa/ecdh.py
merlokk: i thought to return bytes but better will be VerifyingKey object

src/ecdsa/ecdh.py
merlokk: doublechecked. pem and der loads via `string`

merlokk:
done

src/ecdsa/ecdh.py
```diff
@@ -109,107 +132,145 @@ def load_private_key(self, private_key):
...
 
-        :param private_key: binary string with the DER encoding of private ECDSA key
-        :return: public (verifying) key from this private key
+        :param private_key_der: string with the DER encoding of private ECDSA key
+        :type private_key_der: string
+        :return: public (verifying) key from this private key.
+        :rtype: VerifyingKey object
         """
-        return self.load_private_key(SigningKey.from_der(private_key))
+        return self.load_private_key(SigningKey.from_der(private_key_der))
```
tomato42: wont' it raise `InvalidCurveError` in case the curves are different?

src/ecdsa/ecdh.py
tomato42: that's still unfixed

merlokk: fixed

src/ecdsa/test_ecdh.py
```diff
@@ -259,22 +259,27 @@ def test_ecdh_der():
     assert sharedsecret == unhexlify(gshared_secret)
 
 
+# Exception classes used by run_openssl.
+class RunOpenSslError(Exception):
```
tomato42: what's wrong with subprocess.SubprocessError?

merlokk: because python 2.x have no error there
```>       except subprocess.SubprocessError:
E       AttributeError: 'module' object has no attribute 'SubprocessError'
```

merlokk:
done

PR #151, opened 2019-10-30
Ellipticcurve tests

tomato42:
update the test coverage for the ellipticcurve module too

### DIFF
--- a/src/ecdsa/test_ellipticcurve.py
+++ b/src/ecdsa/test_ellipticcurve.py
@@ -1,111 +1,128 @@
-from .ellipticcurve import CurveFp, INFINITY, Point
+import pytest
 from six import print_
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+from hypothesis import given, settings
+import hypothesis.strategies as st
+try:
+    from hypothesis import HealthCheck
+    HC_PRESENT=True
+except ImportError:
+    HC_PRESENT=False
+from .numbertheory import inverse_mod
+from .ellipticcurve import CurveFp, INFINITY, Point
 
-def test_ellipticcurve():
 
-  class FailedTest(Exception):
-    pass
+HYP_SETTINGS={}
+if HC_PRESENT:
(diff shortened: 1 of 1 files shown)


src/ecdsa/test_ellipticcurve.py
```diff
@@ -1,111 +1,129 @@
...
+if HC_PRESENT:
+    HYP_SETTINGS['suppress_health_check']=[HealthCheck.too_slow]
+    HYP_SETTINGS['deadline'] = 5000
+
+
+# NIST Curve P-192:
+p = 6277101735386680763835789423207666416083908700390324961279
+r = 6277101735386680763835789423176059013767194773182842284081
+# s = 0x3045ae6fc8422f64ed579528d38120eae12196d5L
+c = 0x3099d2bbbfcb2538542dcd5fb078b6ef5f3d6fe2c745de65
```
Jakuje: Is the `c` used anywhere?

src/ecdsa/test_ellipticcurve.py
```diff
@@ -1,111 +1,129 @@
...
+Gx = 0x188da80eb03090f67cbf20eb43a18800f4ff0afd82ff1012
+Gy = 0x07192b95ffc8da78631011ed6b24cdd573f977a11e794811
+
+
+c192 = CurveFp(p, -3, b)
+p192 = Point(c192, Gx, Gy, r)
+
+
+def test_p192():
+
```
Jakuje: nit: I think this empty line does not need to be here.

src/ecdsa/test_ellipticcurve.py
tomato42: no, it's not, will comment it out (since it's part of the formal definition of the curve, it may be useful for future tests)

src/ecdsa/test_ellipticcurve.py
tomato42: true, nice catch

PR #157, opened 2019-11-03
der: fix encoding and decoding OIDs

### LINKED ISSUE #155: incorrect handling of OIDs with large second subidentifier
See [X.690 Section 8.19.5](http://handle.itu.int/11.1002/1000/12483-en?locatt=format:pdf&auth)

the correct encoding of OID (2 999 3) is `b'\x06\x03\x88\x37\x03`

that is, if the first sub identifier is 2, then the encoding of the first number in the OID is always equal to `2 * 40 + second` and it may use more than one byte

The `encode_oid()` method can't handle the second subID bigger than 39 while `remove_object()` decodes it as (26, 39, 3), which is not correct.

side note: the test coverage will need to be updated to generate test cases with second subidentifier > 39 when first sub identifier is == 2

tomato42:
make the code handle oids with a large second subidentifier (fixes #155)

make the code raise correct exception when the encoded multi-byte
subidentifier is missing the last byte

move OID test coverage to test_der module and extend it

update the OID generators in random DER generator to generate OIDs with
large second subidentifier

### DIFF
--- a/src/ecdsa/der.py
+++ b/src/ecdsa/der.py
@@ -219,7 +219,8 @@ def read_number(string):
     llen = 0
     if str_idx_as_int(string, 0) == 0x80:
         raise UnexpectedDER("Non minimal encoding of OID subidentifier")
-    # base-128 big endian, with b7 set in all but the last byte
+    # base-128 big endian, with most significant bit set in all but the last
+    # byte
     while True:
         if llen >= len(string):
             raise UnexpectedDER("ran out of length bytes")
--- a/src/ecdsa/der.py
+++ b/src/ecdsa/der.py
@@ -217,6 +217,8 @@ def remove_integer(string):
 def read_number(string):
     number = 0
     llen = 0
+    if str_idx_as_int(string, 0) == 0x80:
+        raise UnexpectedDER("Non minimal encoding of OID subidentifier")
     # base-128 big endian, with b7 set in all but the last byte
     while True:
         if llen >= len(string):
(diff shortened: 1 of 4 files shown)


src/ecdsa/der.py
```diff
@@ -3,6 +3,7 @@
 import binascii
 import base64
 import warnings
+from itertools import chain
 from six import int2byte, b, text_type
 from ._compat import str_idx_as_int
```
sathit9999-Box: src/ecdsa/der.py

src/ecdsa/der.py
```diff
@@ -97,12 +98,10 @@ def encode_octet_string(s):
...
-                                                    for p in pieces]
-    body = b('').join(encoded_pieces)
-    return b('\x06') + encode_length(len(body)) + body
+    assert 0 <= first < 2 and 0 <= second <= 39 or first == 2 and 0 <= second
+    body = b''.join(chain([encode_number(40*first+second)],
+                          (encode_number(p) for p in pieces)))
+    return b'\x06' + encode_length(len(body)) + body
 
 
 def encode_sequence(*encoded_pieces):
```
sathit9999-Box: src/ecdsa/der.py

sathit9999-Box: yes

sathit9999-Box: yes

sathit9999-Box:
yes

PR #161, opened 2019-11-15
Added __eq__ method to classes to improve comparison abilities

psechenov:


src/ecdsa/ellipticcurve.py
```diff
@@ -178,7 +185,7 @@ def double(self):
     x3 = (l * l - 2 * self.__x) % p
     y3 = (l * (self.__x - x3) - self.__y) % p
 
-    return Point(self.__curve, x3, y3)
+    return Point(self.__curve, x3, y3, self.__order)
```
tomato42: the order is not passed here intentionally as it makes the point creation very slow (as the order is then validated)

same in __add__ and __neg__

src/ecdsa/ecdsa.py
```diff
@@ -119,6 +119,13 @@ def __init__(self, generator, point):
     if point.x() < 0 or n <= point.x() or point.y() < 0 or n <= point.y():
       raise RuntimeError("Generator point has x or y out of range.")
 
+  def __eq__(self, other):
+    if isinstance(other, Public_key):    
+      """Return True if the points are identical, False otherwise."""
+      return self.curve == other.curve \
+        and self.point == other.point
+    return False
```
tomato42: I wonder if returning `NotImplemented`, like here:
https://github.com/warner/python-ecdsa/pull/127/files#diff-54eaad15d2c52f2d700a80792692c6b1L80-L82

wouldn't be a better idea; then alternative implementation could handle it, irrespective if the order of operands

src/ecdsa/keys.py
```diff
@@ -649,6 +656,15 @@ def __init__(self, _error__please_use_generate=None):
         self.baselen = None
         self.verifying_key = None
         self.privkey = None
+        
+    def __eq__(self, other):
+        """Return True if the points are identical, False otherwise."""
```
tomato42: it's talking about points but is also checking `default_hashfunc`...

tomato42: also, I wonder if taking default_hashfunc into account is correct at all

src/ecdsa/ecdsa.py
psechenov: At least it makes more sense than just return False

src/ecdsa/keys.py
psechenov: yeah, hash here is more about implementation than instance of the key

src/ecdsa/ellipticcurve.py
```diff
@@ -45,6 +45,14 @@ def __init__(self, p, a, b):
     self.__p = p
     self.__a = a
     self.__b = b
+    
+  def __eq__(self, other):
+    if isinstance(other, CurveFp):    
+      """Return True if the points are identical, False otherwise."""
```
tomato42: points, not curves?

src/ecdsa/test_ecdsa.py
```diff
@@ -386,3 +437,5 @@ def test_sig_verify(args):
     assert pubkey.verifies(msg, signature)
 
     assert not pubkey.verifies(msg - 1, signature)
+
+
```
tomato42: nit: whitespace at the end of file

src/ecdsa/test_ecdsa.py
```diff
@@ -60,6 +60,57 @@ def test_rejection(self):
         assert not self.pubk.verifies(self.msg - 1, self.sig)
 
 
+class TestPublicKey(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        pass
```
tomato42: why define it if it's unused?

src/ecdsa/test_ecdsa.py
```diff
@@ -60,6 +60,57 @@ def test_rejection(self):
...
+        x1 = 0xc58d61f88d905293bcd4cd0080bcb1b7f811f2ffa41979f6
+        y1 = 0x8804dc7a7c4c7f8b5d437f5156f3312ca7d6de8a0e11867f  
+        point1 = ellipticcurve.Point(gen.curve(), x1, y1)
+                
+        x2 = 0x6a223d00bd22c52833409a163e057e5b5da1def2a197dd15
+        y2 = 0x7b482604199367f1f303f9ef627f922f97023e90eae08abf
+        point2 = ellipticcurve.Point(gen.curve(), x2, y2)
+        
+        pub_key1 = Public_key(gen, point1)
+        pub_key2 = Public_key(gen, point2)
```
tomato42: it's testing `Public_key` object defined in `ecdsa` module, shouldn't this test case be in `test_ecdsa.py` then?

also, it would be nice to have test coverage for the NotImplemented lines, comparing those object with `None` or a string should be enough

src/ecdsa/test_ellipticcurve.py
```diff
@@ -126,3 +126,34 @@ def test_multiply(c, x1, y1, m, x3, y3):
...
+        cls.g_23 = Point(cls.c_23, 13, 7, 7)
+        
+    def test_equality_points(self):
+        self.assertEqual(self.g_23, Point(self.c_23, 13, 7, 7))
+        
+    def test_inequality_points(self):
+        c = CurveFp(100, -3, 100)
+        p = Point(c, 100, 100, 100)
+        self.assertNotEqual(self.g_23, p)
+
```
tomato42: nit: whitespace at end of file

src/ecdsa/test_ellipticcurve.py
```diff
@@ -126,3 +126,34 @@ def test_multiply(c, x1, y1, m, x3, y3):
...
+    @classmethod
+    def setUpClass(cls):
+        cls.c_23 = CurveFp(23, 1, 1)
+        
+    def test_equality_curves(self):
+        self.assertEqual(self.c_23, CurveFp(23, 1, 1))
+        
+    def test_inequality_curves(self):
+        c192 = CurveFp(p, -3, b)
```
tomato42: the other test cases use global variables because they use the functional style and the pytest framework, this one uses the object-based approach from unittest, so I'd say that the use of global variables is less clean here...

src/ecdsa/test_ellipticcurve.py
```diff
@@ -126,3 +126,34 @@ def test_multiply(c, x1, y1, m, x3, y3):
...
+        cls.c_23 = CurveFp(23, 1, 1)
+        cls.g_23 = Point(cls.c_23, 13, 7, 7)
+        
+    def test_equality_points(self):
+        self.assertEqual(self.g_23, Point(self.c_23, 13, 7, 7))
+        
+    def test_inequality_points(self):
+        c = CurveFp(100, -3, 100)
+        p = Point(c, 100, 100, 100)
+        self.assertNotEqual(self.g_23, p)
```
tomato42: inequality checks with object of different type missing

tomato42:
side note: it looks like you have not associated the email you used in commits with your github account, you may want to add it Settings→Emails

src/ecdsa/test_ellipticcurve.py
psechenov: Moved to classes. Not sure if it is possible to convert test_p192_mult_tests preserving functionality.

tomato42: I mean, I'm fine with either approach, what I'm not a fan of is mixing them

PR #166, opened 2019-11-30
add vectors from RFC4754

### LINKED ISSUE #5: add Known-Answer-Tests from RFC4754
http://tools.ietf.org/html/rfc4754#page-5

They might already be in ecdsa.py, it has a lot of tests.

tomato42:
fixes #5


src/ecdsa/test_pyecdsa.py
```diff
@@ -1339,3 +1339,68 @@ def test_brainpoolP512r1(self):
             y_Z=int("7DB71C3DEF63212841C463E881BDCF055523BD368240E6C3143BD8DEF8"
                     "B3B3223B95E0F53082FF5E412F4222537A43DF1C6D25729DDB51620A83"
                     "2BE6A26680A2", 16))
+
```
sahanaprasad07: extra line

tomato42: https://www.python.org/dev/peps/pep-0008/: "Surround top-level function and class definitions with two blank lines."

PR #168, opened 2019-11-30
Large hashes with small curves

### LINKED ISSUE #129: Support large hashes with small curves
Current code will abort if the order of the curve is significantly smaller than the size of provided hash. I.e. it's not possible to use SHA-512 with NIST256p.

While it's not a recommended configuration, it is well-defined, and actually used in popular protocols (e.g. TLSv1.2).

Add an option to verify() and sign() methods to change if they will accept bigger inputs and appropriately truncate them.

tomato42:
fixes #129

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -589,8 +589,6 @@ class VerifyingKey(object):
         :type sigdecode: callable
 
         :raises BadSignatureError: if the signature is invalid or malformed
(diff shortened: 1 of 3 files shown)


tomato42:
It's specified in X9.62 but it's not freely available – the fact that it interoperates with openssl will need to suffice.

I wasn't able to find either RFC or NIST SP document that would explicitly state the procedure (both redirect to X9.62)

PR #182, opened 2020-01-16
Fix for Python 3.10, 4 and 10

hugovk:
It's generally safer to compare to the `sys.version_info` tuple than the `sys.version` string, which can include assumptions that each version component is always a single character.

---

There's also some code which uses `six.PY3`, similar to:

```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code" 
```

Where:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code!

Instead, use `six.PY2`.

---

Found with https://github.com/asottile/flake8-2020.

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -68,7 +68,7 @@ Primary classes for performing signing and verification operations.
 
 import binascii
 from hashlib import sha1
-from six import PY3, b
+from six import PY2, b
 from . import ecdsa
 from . import der
 from . import rfc6979
@@ -823,7 +823,7 @@ class SigningKey(object):
         """
         # the privkey pem may have multiple sections, commonly it also has
         # "EC PARAMETERS", we need just "EC PRIVATE KEY".
-        if PY3 and isinstance(string, str):
+        if not PY2 and isinstance(string, str):
             string = string.encode()
         privkey_pem = string[string.index(b("-----BEGIN EC PRIVATE KEY-----")):]
         return cls.from_der(der.unpem(privkey_pem), hashfunc)
--- a/src/ecdsa/util.py
+++ b/src/ecdsa/util.py
@@ -5,7 +5,7 @@ import math
 import binascii
 import sys
 from hashlib import sha256
-from six import PY3, int2byte, b, next
+from six import PY2, int2byte, b, next
 from . import der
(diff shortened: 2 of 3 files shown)


hugovk:
The Python 3.4 failure is unrelated and also happens on `master`, because it's EOL and dependencies are dropping support for it:

* https://travis-ci.org/hugovk/python-ecdsa/builds/637634120

src/ecdsa/keys.py
```diff
@@ -823,7 +823,7 @@ def from_pem(cls, string, hashfunc=sha1):
         """
         # the privkey pem may have multiple sections, commonly it also has
         # "EC PARAMETERS", we need just "EC PRIVATE KEY".
-        if PY3 and isinstance(string, str):
+        if not PY2 and isinstance(string, str):
```
tomato42: why `six.PY3` is unsafe?

src/ecdsa/util.py
```diff
@@ -17,7 +17,7 @@
 oid_ecPublicKey = (1, 2, 840, 10045, 2, 1)
 encoded_oid_ecPublicKey = der.encode_oid(*oid_ecPublicKey)
 
-if sys.version > '3':
+if sys.version_info >= (3,):
```
tomato42: good catch!

src/ecdsa/keys.py
hugovk:
```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code" 
```

Where in six.py:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code!

Instead, use `six.PY2`.

hugovk:
> we are targeting all python versions that have support, be it from community in Debian or from Red Hat in RHEL, we don't limit ourselves to upstream python support

Sure, I guess you'll need to pin some more dependencies.

tomato42:
ok, builds should be fixed with https://github.com/warner/python-ecdsa/pull/183, could you rebase on current master?

hugovk:
Done and Travis is ✅!

hugovk:
You're welcome, thank you for maintaining ecdsa!

PR #184, opened 2020-01-17
Fix Return Type For to_pem 

munagekar:
```python
>>> import ecdsa
>>> private_key = ecdsa.SigningKey.generate(curve=ecdsa.NIST256p)
>>> pub_key_pem_bytes = private_key.get_verifying_key().to_pem()
>>> type(pub_key_pem_bytes)
<class 'bytes'>
```

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -531,7 +531,10 @@ class VerifyingKey(object):
             implementations, it is as big as "uncompressed".
 
         :return: portable encoding of the public key
-        :rtype: str
+        :rtype: bytes
+
+        .. warning:: The PEM is encoded to US-ASCII, it needs to be
+            re-encoded if the system is incompatible (e.g. uses UTF-16)
         """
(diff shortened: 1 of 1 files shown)


tomato42:
yes, the documentation indeed is incorrect

that being said, the PEM encoding should be `str` on python3 as both the headers as well as the base64 encoding inside are supposed to use human-readable characters (i.e. when PEM specifies "A" it means Unicode 0x0041 LATIN CAPITAL LETTER A, _not_ a single byte with value 65)

fixing it would break API so it's not something I'd like to do, so I guess we need to live with it, but I'd prefer to see a warning in the doc string that the PEM byte string returned is already encoded to US-ASCII so it needs to be re-encoded if the system or protocol in which it is used in is not compatible with US-ASCII (e.g. uses UTF-16 or EBCDIC)

munagekar:
Thank you for the review. I have added a warning.

tomato42:
thanks!

Issue #171, opened 2019-11-30
unify formatting between modules


tomato42:
currently modules use formatting all over the place, some use two spaces, some use the standard 4 spaces

remove the `E111` and `E114` from `flake8` configuration in `tox.ini`, fix the issues indentified

(depends on #170 )

hukkinj1:
How about running https://github.com/psf/black once, and enforcing it in CI?

tomato42:
that could also be an option, two things:
1. I don't like the 88 char limit, I'd prefer to stick to 80 (those 8 chars do make a difference on side-by-side diffs)
2. I worry about  py2.6 compatibility, there are some new syntax things in py2.7 that are missing from py2.6

if there's no problem with those, then I have no issues using it and enforcing in CI

hukkinj1:
Here's the PR https://github.com/warner/python-ecdsa/pull/186

> 1. I don't like the 88 char limit, I'd prefer to stick to 80 (those 8 chars do make a difference on side-by-side diffs)

I ran black with max line length set to 79, which is the flake8 default that was used in this project.

> 2. I worry about py2.6 compatibility, there are some new syntax things in py2.7 that are missing from py2.6

Py2.6 tests seem to all pass, so if we can trust them, then all should be good. Perhaps it could also be a good time to drop 2.6 (or all of python 2) support now that even 2.7 was EOL'd recently?

PR #186, opened 2020-01-26
Format with black

hukkinj1:
- Edited tox.ini:
    - Format every python file with black (pinned to version 19.10b0)
    - Use line length 79 for black (this is the flake8 default).
- Ran `black --line-length 79 .` to auto-format every python file

This automatically fixes flake8 errors of types E111,E114,E226,E231,E266,E302,E305,E502,W391.

Possible closes https://github.com/warner/python-ecdsa/issues/171


hukkinj1:
Two questions @tomato42:
1. You mentioned in the issue a line length of 80. I used 79 here because that is the flake8 setting currently in use. I can change to 80 chars if you want?
2. I set the black configuration (i.e. the altered line length) as a CLI arg in `tox.ini`. An alternative is creating a `pyproject.toml` file and adding the configuration there. Downside: one more conf file, upside: the conf is applied when calling black outside tox.ini (e.g. IDE autoformatter, pre-commit hooks etc.).

tox.ini
```diff
@@ -74,24 +74,16 @@ commands = {envpython} speed.py
...
 # We're just getting started. For now, ignore the following problems:
-# E111: indentation is not a multiple of four
-# E114: indentation is not a multiple of four (comment)
-# E226: missing whitespace around arithmetic operator
-# E231: missing whitespace after ','
-# E266: too many leading '#' for block comment
-# E302: expected 2 blank lines, found 1
-# E305: expected 2 blank lines after class or function definition, found 1
+# E203: whitespace before ':' (this needs to be ignored for black compatibility)
 # E501: line too long
```
tomato42: shouldn't we be able to remove this one too?

hukkinj1: Theres a bunch of lines like
```
_b = 0x051953EB9618E1C9A1F929A21A0B68540EEA2DA725B99B315F3B8B489918EF109E156193951EC7E937B1652C0BD3BB1BF073573DF883D2C34F1EF451FD46B503F00
```
which is something black will not fix. These need to be manually sliced or we have to add `# noqa: E501` comments on these lines (personally would prefer the latter). This could also be a separate MR?

tomato42: those can be translated to `int("0x0519...", 16)` this is the approach we are already using for a lot of constants like this

I guess we can do it as a separate PR...

hukkinj1:
Do you perhaps mean flake8? Seems to me that this project is not using pylint.

That check is ignored due to lines like the one mentioned here https://github.com/warner/python-ecdsa/pull/186#discussion_r371229997

Autoformatter wont be able to fix these.

tomato42:
ok, let's move E501 to a separate issue

tox.ini
tomato42: moved  to #187

PR #189, opened 2020-03-09
fix rtype for SigningKey factory methods docstrings

alzix:
SigningKey.from_der - returns SigningKey
SigningKey.from_pem - returns SigningKey


tomato42:
good catch! thanks!

PR #195, opened 2020-04-18
Add support for PKCS#8 private keys

### LINKED ISSUE #113: Add support for PKCS#8 file format
The only format supported for private key files is the legacy "sslay" format. The standardised one is actually PKCS#8, but neither `SigningKey.from_pem()` nor `SigningKey.from_der()` are able to process it. Similarly, `to_der()` and `to_pem()` can't create it.

Extend those methods so that they can read and write PKCS#8 files with ECDSA keys

(this is only about the un-encrypted format, support for encrypted files can come later)

see also #119

oyamauchi:
Closes #113

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -848,8 +848,8 @@ class SigningKey(object):
         Initialise from key stored in :term:`PEM` format.
 
         The PEM formats supported are the un-encrypted RFC5915
-        (the sslay format) supported by OpenSSL, and the more common RFC5958
-        (the PKCS #8 format).
+        (the sslay format) supported by OpenSSL, and the more common
+        un-encrypted RFC5958 (the PKCS #8 format).
 
         The legacy format files have the header with the string
         ``BEGIN EC PRIVATE KEY``.
@@ -876,30 +876,23 @@ class SigningKey(object):
         if not PY2 and isinstance(string, str):
             string = string.encode()
 
-        # the privkey pem may have multiple sections, commonly it also has
-        # "EC PARAMETERS", we need just "EC PRIVATE KEY".
-        ec_private_key_index = string.find(b"-----BEGIN EC PRIVATE KEY-----")
-        if ec_private_key_index != -1:
-            return cls.from_der(
-                der.unpem(string[ec_private_key_index:]), hashfunc, pkcs8=False
-            )
+        # The privkey pem may have multiple sections, commonly it also has
+        # "EC PARAMETERS", we need just "EC PRIVATE KEY". PKCS#8 should not
+        # have the "EC PARAMETERS" section; it's just "PRIVATE KEY".
+        private_key_index = string.find(b"-----BEGIN EC PRIVATE KEY-----")
+        if private_key_index == -1:
+            private_key_index = string.index(b"-----BEGIN PRIVATE KEY-----")
 
-        private_key_index = string.find(b"-----BEGIN PRIVATE KEY-----")
-        if private_key_index != -1:
-            return cls.from_der(
-                der.unpem(string[private_key_index:]), hashfunc, pkcs8=True
-            )
-
-        raise ValueError("No EC PRIVATE KEY or PRIVATE KEY section in PEM")
+        return cls.from_der(der.unpem(string[private_key_index:]), hashfunc)
 
     @classmethod
-    def from_der(cls, string, hashfunc=sha1, pkcs8=False):
+    def from_der(cls, string, hashfunc=sha1):
         """
         Initialise from key stored in :term:`DER` format.
 
         The DER formats supported are the un-encrypted RFC5915
-        (the sslay format) supported by OpenSSL, and the more common RFC5958
-        (the PKCS #8 format).
+        (the sslay format) supported by OpenSSL, and the more common
+        un-encrypted RFC5958 (the PKCS #8 format).
 
         Both formats contain an ASN.1 object following the syntax specified
         in RFC5915::
@@ -911,10 +904,14 @@ class SigningKey(object):
               publicKey  [1] BIT STRING OPTIONAL
             }
 
+        `publicKey` field is ignored completely (errors, if any, in it will
+        be undetected).
+
         The only format supported for the `parameters` field is the named
         curve method. Explicit encoding of curve parameters is not supported.
         In the legacy sslay format, this implementation requires the optional
-        `parameters` field to get the curve name.
+        `parameters` field to get the curve name. In PKCS #8 format, the curve
+        is part of the PrivateKeyAlgorithmIdentifier.
 
         The PKCS #8 format includes this object as the `privateKey` field
         within a larger structure:
@@ -929,14 +926,11 @@ class SigningKey(object):
                 ...
             }
 
-        `publicKey` field is ignored completely (errors, if any, in it will
-        be undetected).
+        The `attributes` and `publicKey` fields are completely ignored; errors
+        in them will not be detected.
 
         :param string: binary string with DER-encoded private ECDSA key
         :type string: bytes like object
-
-        :param pkcs8: whether to expect the data in PKCS #8 format
-        :type pkcs8: boolean
 
         :raises MalformedPointError: if the length of encoding doesn't match
             the provided curve or the encoded values is too large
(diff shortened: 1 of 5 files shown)


src/ecdsa/keys.py
```diff
@@ -870,30 +873,36 @@ def from_pem(cls, string, hashfunc=sha1):
...
+        if ec_private_key_index != -1:
+            return cls.from_der(
+                der.unpem(string[ec_private_key_index:]), hashfunc, pkcs8=False
+            )
+
+        private_key_index = string.find(b"-----BEGIN PRIVATE KEY-----")
+        if private_key_index != -1:
+            return cls.from_der(
+                der.unpem(string[private_key_index:]), hashfunc, pkcs8=True
+            )
```
tomato42, line 891: that feels to me like a lot of code duplication just to set a single variable

src/ecdsa/keys.py
```diff
@@ -904,16 +913,31 @@ def from_der(cls, string, hashfunc=sha1):
...
 
         `publicKey` field is ignored completely (errors, if any, in it will
         be undetected).
 
         :param string: binary string with DER-encoded private ECDSA key
         :type string: bytes like object
 
+        :param pkcs8: whether to expect the data in PKCS #8 format
+        :type pkcs8: boolean
+
```
tomato42, line 940: why force the user to tell use what's the format of the file when we can detect it?

I mean, for most users the different private kye formats are completely arcane, I'd say we should be lenient here, at least by default, in what we accept.

there could be an optional parameter that forces a particular format and aborts reads if the file doesn't match, but that's more of an expert option, with limited usability.

src/ecdsa/keys.py
```diff
@@ -870,30 +873,36 @@ def from_pem(cls, string, hashfunc=sha1):
...
         """
         Initialise from key stored in :term:`DER` format.
 
-        Note, the only DER format supported is the RFC5915
-        (the sslay format) supported by OpenSSL, the more common PKCS#8 format
-        is NOT supported (see:
-        https://github.com/warner/python-ecdsa/issues/113 )
+        The DER formats supported are the un-encrypted RFC5915
+        (the sslay format) supported by OpenSSL, and the more common RFC5958
+        (the PKCS #8 format).
```
tomato42, line 902: the PKCS#8 format also supports encryption, something this code does not, the documentation should reflect that

src/ecdsa/keys.py
```diff
@@ -904,16 +913,31 @@ def from_der(cls, string, hashfunc=sha1):
...
         The only format supported for the `parameters` field is the named
         curve method. Explicit encoding of curve parameters is not supported.
-
-        While `parameters` field is defined as optional, this implementation
-        requires its presence for correct parsing of the keys.
+        In the legacy sslay format, this implementation requires the optional
+        `parameters` field to get the curve name.
+
+        The PKCS #8 format includes this object as the `privateKey` field
+        within a larger structure:
```
tomato42, line 920: a note where the curve is stored in PKCS#8 format is missing

src/ecdsa/keys.py
```diff
@@ -923,8 +947,48 @@ def from_der(cls, string, hashfunc=sha1):
...
+        if pkcs8:
+            s, empty = der.remove_sequence(s)
+            if empty != b(""):
+                raise der.UnexpectedDER(
+                    "trailing junk after DER privkey: %s"
+                    % binascii.hexlify(empty)
+                )
+
+            version, s = der.remove_integer(s)
+            if version != 0 and version != 1:
```
tomato42, line 965: `version not in (0, 1)` would be terser, wouldn't it?

src/ecdsa/keys.py
```diff
@@ -923,8 +947,48 @@ def from_der(cls, string, hashfunc=sha1):
...
+                raise der.UnexpectedDER(
+                    "expected '0' or '1' at start of DER privkey, got %d"
```
tomato42, line 967: why not say that the encoded version is unsupported?

src/ecdsa/keys.py
tomato42: also, I wonder if we shouldn't at least check if presence of `publicKey` matches the `version` (even if we don't parse `publicKey`)

src/ecdsa/keys.py
```diff
@@ -923,8 +947,48 @@ def from_der(cls, string, hashfunc=sha1):
...
+                raise der.UnexpectedDER(
+                    "unexpected algorithm identifier '%s'" % algorithm_oid
+                )
+            if empty != b"":
+                raise der.UnexpectedDER(
+                    "unexpected data after algorithm identifier: %s"
+                    % binascii.hexlify(empty)
+                )
+
+            # We don't care about the optional fields after the key data.
```
tomato42, line 988: 1. why?
2. shouldn't it be documented in the doc text?

tomato42:
the branches identified by coveralls as missing coverage should also be unit tested

src/ecdsa/keys.py
oyamauchi: This will be cleared up once `from_der` detects the incoming format

src/ecdsa/keys.py
oyamauchi: How do you suggest detecting the format, given only DER-encoded data? Both formats start with an integer; if it's `0` we know it's PKCS#1 but if it's `1` it could be either. Should it just detect whether the next thing after the version is an octet-string (PKCS#1) or a sequence (PKCS#8)? That feels brittle to me, but may be good enough.

src/ecdsa/keys.py
oyamauchi: Not sure what you mean here... the version only indicates whether the public key is present in the top-level structure.

src/ecdsa/keys.py
oyamauchi: The optional fields are `attributes` (public key metadata, like certificates) and `publicKey`. They're show in the docstring; I can be explicit about why they're not needed.

src/ecdsa/keys.py
oyamauchi: I feel like that wouldn't be worthwhile, but I could do it if you feel strongly.

src/ecdsa/keys.py
oyamauchi: Will do

src/ecdsa/keys.py
tomato42: yes, detection of format of ASN.1 objects works by checking which description it matches, so looking for a sequence instead of octet-string is the correct approach

src/ecdsa/keys.py
tomato42: regarding `version` or `publicKey`? :)
I don't feel strongly about the latter, the former looks like more readable code to me, so would insist on it

src/ecdsa/keys.py
tomato42: but the element of the structure is called "version", so we should tell the user that the decoded "version" is not one we expect, and it's X

src/ecdsa/keys.py
tomato42: ok, please add comment what we're skipping here and a generic addition in doc string that only such and such fields are inspected (or such and such are completely ignored)

src/ecdsa/keys.py
tomato42: yes and no, I think we still should abort on `BEGIN RSA PRIVATE KEY` and such
also if we have option to force the format, not only the inside DER but the outside PEM needs to match the expected format

oyamauchi:
I don't see how my changes could have caused this decrease in code coverage (the python3.3 version of `hmac_compat` not getting called).

src/ecdsa/keys.py
```diff
@@ -902,14 +904,30 @@ def from_der(cls, string, hashfunc=sha1):
...
+        `publicKey` field is ignored completely (errors, if any, in it will
+        be undetected).
+
         The only format supported for the `parameters` field is the named
         curve method. Explicit encoding of curve parameters is not supported.
+        In the legacy sslay format, this implementation requires the optional
+        `parameters` field to get the curve name. In PKCS #8 format, the curve
+        is part of the PrivateKeyAlgorithmIdentifier.
+
+        The PKCS #8 format includes this object as the `privateKey` field
```
tomato42, line 916: using `this` for subject of a sentence two paragraphs before feels a bit vague to me

src/ecdsa/keys.py
```diff
@@ -981,7 +1049,7 @@ def to_string(self):
         s = number_to_string(secexp, self.privkey.order)
         return s
 
-    def to_pem(self, point_encoding="uncompressed"):
+    def to_pem(self, point_encoding="uncompressed", format="sslay"):
```
tomato42, line 1052: sorry, it looks like I made a typo when documenting the methods, the format is "ssleay", from the predecessor of openssl: https://en.wikipedia.org/wiki/SSLeay

src/ecdsa/keys.py
```diff
@@ -990,9 +1058,11 @@ def to_pem(self, point_encoding="uncompressed"):
         Only the named curve format is supported.
         The public key will be included in generated string.
 
-        The PEM header will specify ``BEGIN EC PRIVATE KEY``
+        The PEM header will specify ``BEGIN EC PRIVATE KEY`` or
+        ``BEGIN PRIVATE KEY``, depending on the desired format.
 
         :param str point_encoding: format to use for encoding public point
+        :param str format: either `sslay` or `pkcs8`
```
tomato42, line 1065:
```
, `ssleay` by default
```

tomato42: btw, if you want `pkcs8` in monospace font, it should use double backticks:
```
``ssleay`` or ``pkcs8``
```

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

src/ecdsa/keys.py
```diff
@@ -1013,6 +1085,7 @@ def to_der(self, point_encoding="uncompressed"):
         The public key will be included in the generated string.
 
         :param str point_encoding: format to use for encoding public point
+        :param str format: either `sslay` or `pkcs8`
```
tomato42, line 1088:
```
, `ssleay` by default
```

tomato42:
awesome, thank you!

PR #199, opened 2020-07-21
🚫🥒 do not pickle the unpickleable RWLock

adhintz:
This is to fix the `can't pickle thread.lock objects` error when attempting to pickle something such as an `ecdsa.keys.VerifyingKey`

Alternatively, the custom getstate and setstate could be within the RWLock class.


src/ecdsa/ellipticcurve.py
```diff
@@ -185,6 +185,15 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
                 doubler = doubler.double().scale()
                 self.__precompute.append((doubler.x(), doubler.y()))
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
```
tomato42, line 189: the RWLock is used to ensure that that the object is self-consistent, this copy should use it

tomato42:
also `black` seems to complain about formatting, please fix that too

adhintz:
Thanks for the response! While writing the test, I cannot get equality comparisons working for PointJacobi objects. Debugging `__eq__` in PointJacobi, it's deciding two PointJacobi objects with the same values are not equal because of `if self.__curve != other.curve():` Although `class CurveFp` defines `__eq__`, it appears to be doing an identity comparison and not using the CurveFp `__eq__`. Here's an example test case:
```
self = <ecdsa.test_jacobi.TestJacobi testMethod=test_can_pickle>

    def test_can_pickle(self):
        pj1 = PointJacobi(curve=CurveFp(23, 1, 1, 1), x=2, y=3, z=1, order=1)
        pj2 = PointJacobi(curve=CurveFp(23, 1, 1, 1), x=2, y=3, z=1, order=1)
>       self.assertEqual(pj1, pj2)
E       AssertionError: <ecdsa.ellipticcurve.PointJacobi object at 0x7f4a0260f0d0> != <ecdsa.ellipticcurve.PointJacobi object at 0x7f4a0260f250>
```
Any idea what I'm doing wrong?

adhintz:
I figured out my question on the equality comparison of PointJacobi. CurveFp has `__eq__` defined, but not `__ne__`. I'll add this to class CurveFp:
```
    def __ne__(self, other):
        return not self.__eq__(other)
```

src/ecdsa/ellipticcurve.py
adhintz: Done. If it can't acquire the lock, I have it return None. Think that's fine or prefer something else?

src/ecdsa/ellipticcurve.py
```diff
@@ -92,14 +92,16 @@ def __init__(self, p, a, b, h=None):
...
 
     def __eq__(self, other):
-        if isinstance(other, CurveFp):
-            """Return True if the curves are identical, False otherwise."""
-            return (
-                self.__p == other.__p
-                and self.__a == other.__a
-                and self.__b == other.__b
-            )
-        return NotImplemented
```
tomato42, line 102: it's intentionally returning `NotImplemented` in case the other is not CurveFp, as then python will execute `__eq__` on `other`, which in turn can have support to coerce CurveFp into its format and correctly compare them (something we actually had to do with Point and PointJacobi already)

src/ecdsa/ellipticcurve.py
```diff
@@ -92,14 +92,16 @@ def __init__(self, p, a, b, h=None):
...
+        """Return True if the curves are identical, False otherwise."""
+        return (
+            isinstance(other, CurveFp)
+            and self.__p == other.__p
+            and self.__a == other.__a
+            and self.__b == other.__b
+        )
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
tomato42, line 104: that special handling of `NotImplemented` also means that `__ne__` should be implemented as `not self == other` (then we also don't have to check for class of other, as the `==` will do that and handle NotImplemented correctly)

src/ecdsa/ellipticcurve.py
```diff
@@ -186,8 +188,14 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
                 self.__precompute.append((doubler.x(), doubler.y()))
 
     def __getstate__(self):
-        state = self.__dict__.copy()
-        del state['_scale_lock']
+        state = None
```
tomato42, line 191: > "When failing to acquire the lock, it now returns None."

I don't think it's possible to fail to acquire lock, it's a blocking operation, and even if it will fail, it will raise an exception, so `return state` won't be reached... as such I don't think `state = None` is necessary

src/ecdsa/ellipticcurve.py
adhintz: Makes sense. Reverted to the original code.

src/ecdsa/ellipticcurve.py
adhintz: Good call! Thanks for catching that.

src/ecdsa/ellipticcurve.py
adhintz: Makes sense. Removed.

PR #208, opened 2020-08-25
Fix too long lines

### LINKED ISSUE #187: Reformat long lines to fit in 79 characters
Some lines are still too long after #186 so we should reformat it so that E501 can be removed from tox.ini

tomato42:
Format code so that it is less than 80 columns.

Fixes #187


ep69 (approved):
Looks good to me.

Issue #180, opened 2020-01-07
Importing anything from ecdsa is too expensive


jhance:
Since all the curves are constructed at import time in the modules __init__ using anything from ecdsa has a substantial import cost (like 0.3-0.4s which is pretty substantial). Computation at import time should be avoided.

tomato42:
> Computation at import time should be avoided.

why? this way the cost is incurred once per application start

jgphilpott:
I think I can see the case for avoiding computation at import... maybe I want to import ecdsa but not use any of its features until some event in the future triggered by human input. Doing the computation at the time of the event rather than at import could save time at application start. Personally I don't feel like 0.3-0.4s is a big deal but I guess it depends how time sensitive your application is.. is that what you're getting at @jhance ?

tomato42:
the calculation that causes this is most likely the one that precomputes the multiplication table for curve generators, here:
https://github.com/warner/python-ecdsa/blob/93b04ba3ddb7c2716e07761393a179c061718c34/src/ecdsa/ellipticcurve.py#L163-L174

triggered by passing in `generator=True` here, and in the rest of that module:
https://github.com/warner/python-ecdsa/blob/93b04ba3ddb7c2716e07761393a179c061718c34/src/ecdsa/ecdsa.py#L296-L297

now, while it could be possible to compute that multiplication table once, and pass in as parameter to init instead of computing it, it would also force us to keep using this simplest possible precomputation table – something I'd rather not do (it's also the reason why `__precompute` is private)

the clever thing to do would be to generate those multiplication tables the first time the generator is actually used, but that is not inherently thread-safe (or easy to make thread-safe given our API)

finally, as far as performance goes, on my machine, import takes about 131ms:
```python
>>> a = time.time(); import ecdsa; b = time.time()
>>> b - a
0.13128066062927246
```
with the old 0.13 version a single keygen operation with smallest curve takes 30ms:
```
 NIST192p: siglen= 48, keygen=0.030s, sign=0.011s, verify=0.022s
 NIST224p: siglen= 56, keygen=0.041s, sign=0.015s, verify=0.030s
 NIST256p: siglen= 64, keygen=0.055s, sign=0.020s, verify=0.040s
SECP256k1: siglen= 64, keygen=0.055s, sign=0.020s, verify=0.039s
 NIST384p: siglen= 96, keygen=0.126s, sign=0.047s, verify=0.093s
[...]

PR #212, opened 2020-11-08
add option to lazily precompute multiplication tables

oberstet:
this is an attempt to fix https://github.com/warner/python-ecdsa/issues/180

the does not change any existing behavior, but adds new behavior:

when `PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE` env var is set, the multiplication tables for curves are only precomputed upon first use.

without the env var being set (the current behavior), tables for _all_ supported curves are precomputed at import time regardless of whether the curves are used in the user app at all. 

effect can be seen in this test:

```
(cpy382_2) oberstet@intel-nuci7:~/scm/3rdparty/python-ecdsa$ python test.py 
python-ecdsa v0.16.0+1.g3388afc
0.4552568449998944
(cpy382_2) oberstet@intel-nuci7:~/scm/3rdparty/python-ecdsa$ PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE=1 python test.py 
python-ecdsa v0.16.0+1.g3388afc
0.020854599999438506
(cpy382_2) oberstet@intel-nuci7:~/scm/3rdparty/python-ecdsa$ 
```

using `test.py`:

```python
import timeit

def test1():
    from ecdsa.curves import SECP256k1

print(timeit.timeit(test1, number=1))
```

### DIFF
--- a/src/ecdsa/__init__.py
+++ b/src/ecdsa/__init__.py
@@ -44,6 +44,7 @@ __all__ = [
     "numbertheory",
     "test_pyecdsa",
     "util",
+    "six",
 ]
 
 _hush_pyflakes = [
@@ -73,4 +74,3 @@ _hush_pyflakes = [
     BRAINPOOLP512r1,
 ]
 del _hush_pyflakes
-
--- a/src/ecdsa/__init__.py
+++ b/src/ecdsa/__init__.py
@@ -1,3 +1,10 @@
+import os
+
+# when enabled, multiplication tables are precomputed only when actually needed:
+# that is, the specific curve is actually first used in user code. otherwise, multiplication
+# tables for _all_ curves supported are precomputed at library _import_ time effectively
+ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
+
 from .keys import (
     SigningKey,
     VerifyingKey,
@@ -44,7 +51,6 @@ __all__ = [
     "numbertheory",
     "test_pyecdsa",
     "util",
-    "six",
 ]
 
 _hush_pyflakes = [
@@ -74,3 +80,4 @@ _hush_pyflakes = [
     BRAINPOOLP512r1,
 ]
 del _hush_pyflakes
+
--- a/src/ecdsa/__init__.py
+++ b/src/ecdsa/__init__.py
@@ -1,10 +1,3 @@
-import os
-
-# when enabled, multiplication tables are precomputed only when actually needed:
-# that is, the specific curve is actually first used in user code. otherwise, multiplication
-# tables for _all_ curves supported are precomputed at library _import_ time effectively
-ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
-
 from .keys import (
     SigningKey,
     VerifyingKey,
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -34,7 +34,6 @@
 # Written in 2005 by Peter Pearson and placed in the public domain.
 
 from __future__ import division
-import os
 
 try:
     from gmpy2 import mpz
@@ -181,7 +180,7 @@ class PointJacobi(object):
     def _maybe_precompute(self):
         if self.__generator:
             # since we lack promotion of read-locks to write-locks, we do a
-            # "acquire-read-lock check, acquire-write-lock plus recheck" cycle ..
+            # "acquire-read-lock check, acquire-write-lock plus recheck" cycle
             try:
                 self._scale_lock.reader_acquire()
                 if self.__precompute:
@@ -197,7 +196,9 @@ class PointJacobi(object):
                 assert order
                 i = 1
                 order *= 2
-                doubler = PointJacobi(self.__curve, self.__x, self.__y, self.__z, order)
+                doubler = PointJacobi(
+                    self.__curve, self.__x, self.__y, self.__z, order
+                )
                 order *= 2
                 self.__precompute.append((doubler.x(), doubler.y()))
 
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -34,6 +34,7 @@
 # Written in 2005 by Peter Pearson and placed in the public domain.
 
 from __future__ import division
+import os
 
 try:
     from gmpy2 import mpz
@@ -177,10 +178,11 @@ class PointJacobi(object):
         self.__generator = generator
         self.__precompute = []
 
-        # unless ENABLE_LAZY_PRECOMPUTE is enabled, precompute multiplication
-        # tables _now_ (at ctor/import time)
-        import ecdsa
-        if not ecdsa.ENABLE_LAZY_PRECOMPUTE:
+        # when enabled, multiplication tables are precomputed only when actually needed:
+        # that is, the specific curve is actually first used in user code. otherwise, multiplication
+        # tables for _all_ curves supported are precomputed at library _import_ time effectively
+        ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
+        if not ENABLE_LAZY_PRECOMPUTE:
             self._maybe_precompute()
 
     def _maybe_precompute(self):
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -178,28 +178,25 @@ class PointJacobi(object):
         self.__generator = generator
         self.__precompute = []
 
-        # when enabled, multiplication tables are precomputed only when actually
-        # needed: that is, the specific curve is actually first used in user code.
-        # otherwise, multiplication tables for _all_ curves supported are precomputed
-        # at library _import_ time effectively
-        ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
-        if not ENABLE_LAZY_PRECOMPUTE:
-            self._maybe_precompute()
-
     def _maybe_precompute(self):
         if self.__generator and not self.__precompute:
             order = self.__order
             assert order
-            i = 1
-            order *= 2
-            doubler = PointJacobi(self.__curve, self.__x, self.__y, self.__z, order)
-            order *= 2
-            self.__precompute.append((doubler.x(), doubler.y()))
-
-            while i < order:
-                i *= 2
-                doubler = doubler.double().scale()
+            try:
+                self._scale_lock.writer_acquire()
+                i = 1
+                order *= 2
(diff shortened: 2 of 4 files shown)


src/ecdsa/__init__.py
```diff
@@ -44,7 +44,6 @@
     "numbertheory",
     "test_pyecdsa",
     "util",
-    "six",
```
tomato42, line 47: yes, it's wrong but it's here for backwards compatibility

src/ecdsa/ellipticcurve.py
```diff
@@ -174,12 +175,23 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
...
-        if generator:
+
+        # when enabled, multiplication tables are precomputed only when actually needed:
+        # that is, the specific curve is actually first used in user code. otherwise, multiplication
+        # tables for _all_ curves supported are precomputed at library _import_ time effectively
+        ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
+        if not ENABLE_LAZY_PRECOMPUTE:
+            self._maybe_precompute()
+
+    def _maybe_precompute(self):
```
tomato42, line 188: this is not thread safe, if two threads use the same object it will end up with malformed `__precompute` table; see `scale()`

src/ecdsa/ellipticcurve.py
```diff
@@ -174,12 +175,23 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
...
             self.__z = z
             self.__order = order
+        self.__generator = generator
         self.__precompute = []
-        if generator:
+
+        # when enabled, multiplication tables are precomputed only when actually needed:
+        # that is, the specific curve is actually first used in user code. otherwise, multiplication
+        # tables for _all_ curves supported are precomputed at library _import_ time effectively
+        ENABLE_LAZY_PRECOMPUTE = os.environ.get('PYTHON_ECDSA_ENABLE_LAZY_PRECOMPUTE', None) is not None
```
tomato42, line 184: I don't like the environment variable idea, this only makes the use of the library more complex, and delayed precomputation benefits everybody that doesn't use all curves every time

tomato42 (changes requested):

tomato42:
The CI also failed: https://travis-ci.org/github/warner/python-ecdsa/builds/742220130

src/ecdsa/ellipticcurve.py
oberstet: as far as I see: all call contexts of `maybe_precompute` already have critical sections

oberstet: and the original pre-computation in the ctor doesn't need one .. I think? an object cannot be accessed from 2 threads while it is still being constructed?

src/ecdsa/__init__.py
oberstet: ok, I'll readd it

src/ecdsa/ellipticcurve.py
oberstet: note sure I understand: so should I remove the controlling flag altogether and make it "always lazy"?

tomato42: yes, make it always lazy

src/ecdsa/ellipticcurve.py
tomato42: technically, if you mess around `__new__` and mess with inheritance really badyly, you could end up with two threads in an `__init__` at the same time, but not in regular use, so yes, we didn't need locks before because in practice it's not possible to have two threads in `__init__` at the same time

> as far as I see: all call contexts of maybe_precompute already have critical sections

huh? as long as two threads start execution of the _maybe_precompute at the same time, and one thread checks the if before the second thread runs the first `.append()` both threads will execute the entirety of it

and even if not, then the other thread can think the __precompute list is fully initialised when it's in the process of being initialised, ending up using incomplete table

src/ecdsa/ellipticcurve.py
```diff
@@ -215,6 +229,7 @@ def __eq__(self, other):
         elif isinstance(other, PointJacobi):
             try:
                 other._scale_lock.reader_acquire()
+                other._maybe_precompute()
```
tomato42, line 232: it's not needed here

src/ecdsa/ellipticcurve.py
```diff
@@ -205,6 +218,7 @@ def __eq__(self, other):
         """Compare two points with each-other."""
         try:
             self._scale_lock.reader_acquire()
+            self._maybe_precompute()
```
tomato42, line 221: it's not needed here

src/ecdsa/ellipticcurve.py
```diff
@@ -574,7 +589,9 @@ def __mul__(self, other):
         if self.__order:
             # order*2 as a protection for Minerva
             other = other % (self.__order * 2)
-        if self.__precompute:
+        self._maybe_precompute()
+        other._maybe_precompute()
```
tomato42, line 593: other is an integer here, not a point

oberstet:
ok, pushed adjustments based on your feedback, and tested on cpy and pypy:

* https://gist.github.com/oberstet/905251d53a5168ea8c56996618a3a4cb
* https://gist.github.com/oberstet/c8ed75f1be47f2eacbfccc9b25d5565d

2 notes:

* in the original code, the precomputation was called with constructor x/y whereas in the new code, x/y can be a gmp wrapped x/y. I tested both the non-gmp and gmp tox envs, so that should be good?
* I now added a writer (single I guess) lock within the precompute block. new locks, means new risks for deadlock. also that lock will now happen later on .. lazy (which is the whole point).

src/ecdsa/ellipticcurve.py
```diff
@@ -574,6 +584,7 @@ def __mul__(self, other):
         if self.__order:
             # order*2 as a protection for Minerva
             other = other % (self.__order * 2)
+        self._maybe_precompute()
```
oberstet, line 587: this will escalate the lock from reader to writer! not sure ... is "getstate" supposed to return also the lazy-computed attribute?

tomato42: > this will escalate the lock from reader to writer! not sure ...

I don't see how...

> is "getstate" supposed to return also the lazy-computed attribute?

yes, it should; the pickling is used for multiprocess module for example

src/ecdsa/ellipticcurve.py
```diff
@@ -174,20 +175,29 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
...
-            order *= 2
-            doubler = PointJacobi(curve, x, y, z, order)
-            order *= 2
-            self.__precompute.append((doubler.x(), doubler.y()))
-
-            while i < order:
-                i *= 2
-                doubler = doubler.double().scale()
+            try:
+                self._scale_lock.writer_acquire()
```
tomato42, line 186: and what if multiple threads execute this line at the same time?

I'm also not a fan of overloading the `_scale_lock`, it's used for a different purpose

oberstet: > and what if multiple threads execute this line at the same time?

it will block in https://docs.python.org/3.8/library/threading.html#threading.Lock.acquire

a deadlock would arise if that code blocking would have previously acquired a write lock upon which the precomputing thread would have to wait subsequently. 

anyways, we'll use our fork for now, works

tomato42: what I mean is that both threads will then add values to the precompute list, that's wrong

oberstet: no, because of `not self.__precompute` on line 182. or are you saying that this latter access has to be under the critical section as well

oberstet: .. but that would escalate each and every "maybe_precompute" call, even when already cached. another option would be to optain a read lock already for the initial access in "maybe_precompute", and then escalate that to write lock without releasing before. not sure if the lock wrapper in use even allows that. mmh. locks;) don't use them ..

tomato42:
> ok, pushed adjustments based on your feedback, and tested on cpy and pypy:
> 
>     * https://gist.github.com/oberstet/905251d53a5168ea8c56996618a3a4cb
> 
>     * https://gist.github.com/oberstet/c8ed75f1be47f2eacbfccc9b25d5565d
> 
> 
> 2 notes:
> 
>     * in the original code, the precomputation was called with constructor x/y whereas in the new code, x/y can be a gmp wrapped x/y. I tested both the non-gmp and gmp tox envs, so that should be good?

check the CI results: https://travis-ci.org/github/warner/python-ecdsa/pull_requests
 
>     * I now added a writer (single I guess) lock within the precompute block. new locks, means new risks for deadlock. also that lock will now happen later on .. lazy (which is the whole point).

the solution to deadlocking is not locking in fewer places but to lock in correct places; here we need to acquire reader lock when checking if the precompute is initialised, free it, and then acquire a writer lock if it wasn't initialised; then after getting the writer lock we need to check again for initialisation

oberstet:
> here we need to acquire reader lock when checking if the precompute is initialised, free it, and then acquire a writer lock if it wasn't initialised; then after getting the writer lock we need to check again for initialisation

I've updated the PR which should reflect ^ now ..

src/ecdsa/ellipticcurve.py
```diff
@@ -34,6 +34,7 @@
 # Written in 2005 by Peter Pearson and placed in the public domain.
 
 from __future__ import division
+import os
```
tomato42, line 37: unused import

src/ecdsa/ellipticcurve.py
```diff
@@ -174,20 +175,40 @@ def __init__(self, curve, x, y, z, order=None, generator=False):
...
-            while i < order:
-                i *= 2
-                doubler = doubler.double().scale()
+
+    def _maybe_precompute(self):
+        if self.__generator:
+            # since we lack promotion of read-locks to write-locks, we do a
+            # "acquire-read-lock check, acquire-write-lock plus recheck" cycle ..
+            try:
+                self._scale_lock.reader_acquire()
```
tomato42, line 186: it's `_scale_lock` because it was used for `scale()` but now we also use it for precompute... maybe change its name to `_update_lock`?

tomato42 (changes requested):

tomato42:
also few previous versions failed in CI because of excessive line length, you may want to run `tox -e codechecks` before pushing a new version

oberstet:
thanks for your time and for reviewing! pls merge.

> sidenote: the CI settings on travis for this repo don't allow for unbuilt pushes to the PR supersede others (to CI run requests pile up), and for whatever reasons, a single CI run takes hours.

tomato42:
the builds take hours because travis limited the number of free workers recently: https://www.traviscistatus.com/#month don't know why exactly, unfortunately I don't have access to administer it to fix the dismissal of stale PRs or lack of CI status in PRs either... I'll send an email to Warner tomorrow

tomato42 (approved):
ok, looks good, but I'll wait until the CI finishes before merging

oberstet:
thanks for approving!

> re: travis: ah, right. interesting - news for me. guess we'll run into this soonish as well on our oss stuff ..

tomato42:
merged, thank you!

Issue #211, opened 2020-11-07
Import Very Slow on Raspberry Pi


om26er:
On a Raspberry Pi, a simple import of ecdsa takes a lot of time, that results in very slow start for our software. In some cases, it can take as much as 30 seconds to import below
```
from ecdsa.curves import SECP256k1
```

During the import, the CPU stays at 100%.

We have tried it on multiple operating systems
  * Ubuntu Server 20.10
  * Yocto based custom image as well.

om26er:
* On a cold pycache it takes 30 seconds to import on the Pi.
* On future attempts it stays above 17 seconds for the import.

oberstet:
fwiw, here is a test on x86-64/pypy3:

![Bildschirmfoto von 2020-11-07 20-13-24](https://user-images.githubusercontent.com/233340/98449686-ca8aa800-2135-11eb-91c4-e49f9ce5e2a8.png)

tomato42:
duplicate of #180

PR #213, opened 2020-11-09
Precompute updates

tomato42:
little bit of cleanup after the addition of precomputation code
add CI on python 3.9 as it's released now


tomato42:
CI results: https://travis-ci.org/github/warner/python-ecdsa/builds/742510210

letoams:
Looks good to me. My only question is what the the precompute has a lazy flag at all. Why not just always have it lazy? Is there a use case where being lazy would be slower?

tomato42:
yes, the first operation with lazy precompute will be much slower (like few hundred times slower), so if the code is latency sensitive, it may be too late to do it the first time the key is actually used

letoams:
> yes, the first operation with lazy precompute will be much slower (like few hundred times slower), so if the code is latency sensitive, it may be too late to do it the first time the key is actually used

But is that really a factor ? If performance were that critical, would you use python? Wouldn't this need to use ctypes ?

If you startup a program to do calculations, then you always take this hit anyway before you can "get to work". So the only use case we are talking about is a daemon waiting to do signing operations. Is "a few hundred times slower" meaningful compared to like 20ms of network latency ?

It feels a bit more like feature creep than actual real life use case to me. But having said that, I don't object to any of it.

tomato42:
> > yes, the first operation with lazy precompute will be much slower (like few hundred times slower), so if the code is latency sensitive, it may be too late to do it the first time the key is actually used
> 
> But is that really a factor ? If performance were that critical, would you use python? Wouldn't this need to use ctypes ?

because on very underperforming platforms (like Raspberry Pi) that few hundred times slower translates for few seconds, see #211 and linked issues
 
> If you startup a program to do calculations, then you always take this hit anyway before you can "get to work". So the only use case we are talking about is a daemon waiting to do signing operations. Is "a few hundred times slower" meaningful compared to like 20ms of network latency ?

yes, on a regular PC it's just 20ms so it doesn't matter, but on a phone it may be 2-3 seconds

> It feels a bit more like feature creep than actual real life use case to me. But having said that, I don't object to any of it.

well, yeah, but it's reusing the code we need to library init so... meh

PR #215, opened 2020-11-11
fix truncation of hash input with order bit size not multiple of 8

### LINKED ISSUE #197: Digest truncation should happen according to bitlength not bytelength
FIPS 186.4 states that:
>If the length of the output of the hash function is greater than the bit length of n, then the leftmost n bits of the hash function output block shall be used in any calculation using the hash function output during the generation or verification of a digital signature.

(6.1.1 https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf)

However in python-ecdsa/src/ecdsa/keys.py 686-688, 1359-1361:
>digest = normalise_bytes(digest)
>if allow_truncate:
>       digest = digest[: self.curve.baselen]

truncation is based on the baselen value, which is calculated by python-ecdsa/src/ecdsa/util.py 62-63:
>def orderlen(order):
    return (1 + len("%x" % order)) // 2  # bytes

which returns the length in bytes and not bits as it should. If the bitlength happens to be a multiple of 8 (as usual) then the issue remains hidden.

tomato42:
all curves we have defined, with exception of NIST P-521, have
bit length of order that's multiple of 8, and there are no hashes
that are longer than 512 bits, so NIST P-521 is not affected either,
but do fix this issue in case user provides raw input longer than
512 bits or uses custom curves

fix #197

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -77,7 +77,7 @@ from .curves import NIST192p, find_curve
 from .numbertheory import square_root_mod_prime, SquareRootError
 from .ecdsa import RSZeroError
 from .util import string_to_number, number_to_string, randrange
-from .util import sigencode_string, sigdecode_string
+from .util import sigencode_string, sigdecode_string, bit_length
 from .util import (
     oid_ecPublicKey,
     encoded_oid_ecPublicKey,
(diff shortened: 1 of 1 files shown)


tomato42:
CI run: https://travis-ci.org/github/warner/python-ecdsa/builds/742931782

The-Mule (approved):
Looks good to me.

PR #217, opened 2020-11-20
Exclude version specific coverage

tomato42:
coveralls complains about branches that are executed in one environment only, but if they are version specific, that's expected
so try excluding them


tomato42:
CI run: https://travis-ci.org/github/warner/python-ecdsa/builds/744777383

tomato42:
CI run: https://travis-ci.org/github/warner/python-ecdsa/builds/744894502

PR #220, opened 2020-12-02
use native inverse modulo when available

tomato42:
the builtin pow() can handle negative powers on python 3.8 and later,
so do use it


tomato42:
Ci run: https://travis-ci.org/github/warner/python-ecdsa/builds/747283435

PR #240, opened 2020-12-14
implement lock-less algorithm in PointJacobi

### LINKED ISSUE #238: Killing long computation in IPython with Ctrl+C sometimes leaves the environment inoperable
When the thread is killed by KeyboardInterrupt it may happen while the `RWLock` was handling the locking of the different queues, leaving the PointJacobi inoperable.

The code should handle abort with SIGINT gracefully, in a way that allows object reuse in IPython or in an interactive session.

tomato42:
since both __coords and __precompute have a "final" state, we can easily
implement lock-less handling of those parameters by depending on GIL
(the atomicity of single bytecode instructions)

fixes #238
closes #239 (this in an alternative, simpler approach)

### DIFF
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -49,7 +49,6 @@ except ImportError:  # pragma: no branch
 
 from six import python_2_unicode_compatible
 from . import numbertheory
-from ._rwlock import RWLock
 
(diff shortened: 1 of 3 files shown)


The-Mule (approved):
Everything seems to be sane and tested too. r+

PR #251, opened 2021-04-24
allow limiting point formats, don't accept malformed PEM public files

### LINKED ISSUE #214: Limit acceptable point encoding formats (raw encoding in ASN.1 files)
While methods like https://github.com/tlsfuzzer/python-ecdsa/blob/73f3d3e47a8b48f60b71baa968cfaec0beeb4917/src/ecdsa/ecdh.py#L240 and https://github.com/tlsfuzzer/python-ecdsa/blob/73f3d3e47a8b48f60b71baa968cfaec0beeb4917/src/ecdsa/keys.py#L275-L277 will conveniently accept all valid point encoding formats for a selected curve, this may lead to a not standard-compliant behaviour. For example, in TLS 1.2, unless compressed point format was negotiated before, its use is forbidden; raw format is never correct.

Extend those methods to allow specifying allowed key formats, update
https://github.com/tlsfuzzer/python-ecdsa/blob/73f3d3e47a8b48f60b71baa968cfaec0beeb4917/src/ecdsa/keys.py#L411
to use it (raw encoding is not allowed in ASN.1, we shouldn't accept it).

tomato42:
Allow specifying what point formats are supported when loading public
keys. Limit the point formats when loading PEM and DER public files
to the formats actually allowed in them: uncompressed, compressed,
and hybrid. Previous code would allow raw encoding too.

fixes #214

### DIFF
--- a/src/ecdsa/ecdh.py
+++ b/src/ecdsa/ecdh.py
@@ -216,7 +216,7 @@ class ECDH(object):
 
         :return: public (verifying) key from local private key.
         :rtype: VerifyingKey object
-       """
+        """
         return self.private_key.get_verifying_key()
 
     def load_received_public_key(self, public_key):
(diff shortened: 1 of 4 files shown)


Jakuje (approved):
lgtm

PR #254, opened 2021-05-28
Fix typos

MartinThoma:


NEWS
```diff
@@ -89,11 +89,11 @@ Support for reading and writing private keys in PKCS#8 format.
 
 New API:
 `to_pem` and `to_der` now accept new parameter, `format`, to specify
-the format of the encoded files, either the dafault, legacy "ssleay", or
+the format of the encoded files, either the default, legacy "SSLeay", or
```
tomato42, line 92: `ssleay` is also the value of the parameter, and it is case-sensitive, so could you rather change the quote marks to backticks instead of the case?

tomato42 (commented):
Thank you! And looks good in general, there is one slight factual inaccuracy that it introduces, could you fix it?

NEWS
MartinThoma: Done :-) :+1:

MartinThoma:
@tomato42 I fixed it :-)

tomato42:
I merged it ;-)
Thanks!

PR #256, opened 2021-07-15
EdDSA support

tomato42:
Add support for EdDSA signatures (Ed25519 and Ed448).

Todo:
 - [x] interoperability tests with openssl
 - [x] precomputation support for point multiplication #262
 - [x] profile and optimise code

### DIFF
--- a/src/ecdsa/test_pyecdsa.py
+++ b/src/ecdsa/test_pyecdsa.py
@@ -43,6 +43,8 @@ from .curves import (
     BRAINPOOLP320r1,
     BRAINPOOLP384r1,
     BRAINPOOLP512r1,
+    Ed25519,
+    Ed448,
     curves,
 )
 from .ecdsa import (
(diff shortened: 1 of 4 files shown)


src/ecdsa/test_pyecdsa.py
```diff
@@ -1354,6 +1356,120 @@ def do_test_to_openssl(self, curve, hash_name="SHA1"):
...
+        "ed448" not in OPENSSL_SUPPORTED_TYPES,
+        reason="system openssl does not support signing with Ed448",
+    )
+    def test_to_openssl_ed448(self):
+        return self.do_eddsa_test_to_openssl(Ed448)
+
+    def do_eddsa_test_from_openssl(self, curve):
+        curvename = curve.name
+
+        if os.path.isdir("t"):
```
ep69, line 1426: What about if there is a file called `t`? Do we care about such situations?

ep69 (approved):
r+ with a small question

src/ecdsa/test_pyecdsa.py
tomato42: yes, it's a problem, but it's been like that for the past 10 years for this library (there are other, non EdDSA tests that use the `t` directory too), so I don't think it's a problem

PR #257, opened 2021-07-26
add SHAKE-256 implementation

tomato42:
On earlier pythons we don't have the ability to set the size of
SHAKE-256 output, so we need to use our own implementation.

Context: #256


mahavrila (approved):
Looks good to me.

PR #258, opened 2021-07-26
Add Twisted Edwards curve point arithmetic

tomato42:
To support EdDSA (#256) we need to have an implementation of point arithmetic over twisted Edwards curves.


beldmit (approved):
LGTM

PR #259, opened 2021-07-26
Low level EdDSA

tomato42:
Add the low level implementation of EdDSA

context: #256


shigaraki0 (commented):
Looks good to me.

tomato42:
Thanks for the review!

PR #262, opened 2021-08-13
eddsa: add support for point precomputation

tomato42:
Significantly speed up all EdDSA operations by computing multiplication table of the generator point


src/ecdsa/ellipticcurve.py
```diff
@@ -1324,8 +1327,41 @@ def from_bytes(
...
+        precompute.append((coord_x, coord_y, coord_t))
+
+        while i < order:
+            i *= 2
+            doubler = doubler.double().scale()
+            coord_x, coord_y = doubler.x(), doubler.y()
+            coord_t = coord_x * coord_y % prime
+            precompute.append((coord_x, coord_y, coord_t))
+
+        self.__precompute = precompute
```
t184256, line 1364: Here are two attempts to restructure the do-while awkwardness, with control flow

``` python
        assert self.__order
        precompute = []
        prime = self.__curve.p()
        doubler = PointEdwards(self.__curve, *self.__coords, self.__order * 2)
        i = 1
        while True:
            doubler.scale()
            coord_x, coord_y = doubler.x(), doubler.y()
            coord_t = coord_x * coord_y % prime
            precompute.append((coord_x, coord_y, coord_t))
            if i >= self.__order * 4:
                break
            i *= 2
            doubler = doubler.double()
        self.__precompute = precompute
        # assert self.__order * 8 < 2**len(precompute) < self.__order * 16
```

and with extracting scale-yield-double to a separate generator

``` python
        def double_gen(p):
            while True:
                yield p.scale()
                p = p.double()

        assert self.__order
        prime = self.__curve.p()
        p1 = PointEdwards(self.__curve, *self.__coords, self.__order * 2)
        import itertools
        self.__precompute = list(itertools.islice(
            ((p.x(), p.y(), p.x() * p.y() % prime) for p in double_gen(p1)),
            bit_length(self.__order) + 3
        ))
        # assert self.__order * 8 < 2**len(self.__precompute) < self.__order * 16
```

src/ecdsa/ellipticcurve.py
```diff
@@ -1492,6 +1550,9 @@ def __mul__(self, other):
         if self.__order:
             # order*2 as a protection for Minerva
             other = other % (self.__order * 2)
+        self._maybe_precompute()
+        if self.__precompute:
```
t184256, line 1554: Maybe it's a good idea to return something truthful, like `self.__precompute` itself, and combine these into one.

src/ecdsa/ellipticcurve.py
```diff
@@ -1482,6 +1518,28 @@ def __rmul__(self, other):
         """Multiply point by an integer."""
         return self * other
 
+    def _mul_precompute(self, other):
+        """Multiply point by integer with precomputation table."""
+        X3, Y3, Z3, T3, p, a = 0, 1, 1, 0, self.__curve.p(), self.__curve.a()
+        _add = self._add
+        for X2, Y2, T2 in self.__precompute:
+            if other % 2:
+                if other % 4 >= 2:
```
t184256, line 1527: I find `== 3` less confusing here. Could be even flatter:

``` python
            if r == 1:
                other = (other - 1) // 2
                X3, Y3, Z3 = _add(X3, Y3, Z3, X2, Y2, 1, p)
            elif r == 3:
                other = (other + 1) // 2
                X3, Y3, Z3 = _add(X3, Y3, Z3, X2, -Y2, 1, p)
            else:  # 0 or 2
                other //= 2
```

src/ecdsa/keys.py
```diff
@@ -263,10 +263,20 @@ def precompute(self, lazy=False):
            use (when set to True)
         """
         if isinstance(self.curve.curve, CurveEdTw):
-            return
-        self.pubkey.point = ellipticcurve.PointJacobi.from_affine(
-            self.pubkey.point, True
-        )
+            pt = self.pubkey.point
+            self.pubkey.point = ellipticcurve.PointEdwards(
```
t184256, line 267: Not a fan of precomputing through a setter, but seems like you've made your mind on it.

src/ecdsa/ellipticcurve.py
```diff
@@ -1482,6 +1518,28 @@ def __rmul__(self, other):
         """Multiply point by an integer."""
         return self * other
 
+    def _mul_precompute(self, other):
```
t184256, line 1521: Is it worth it directly cross-testing this against non-precomputing implementation?

t184256 (commented):
Not a fan of some details, though you're unlikely to find them blocking either.

src/ecdsa/eddsa.py
```diff
@@ -116,6 +116,12 @@ def __ne__(self, other):
     def point(self):
         return self.__point
 
+    @point.setter
+    def point(self, other):
+        if self.__point != other:
```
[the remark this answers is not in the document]
tomato42: Point doesn't have assigned generator, so it can't change, changing order of the point is explicitly allowed, I'm pretty sure that `INFINITY` will not be allowed...

src/ecdsa/ellipticcurve.py
```diff
@@ -1284,6 +1284,8 @@ def __init__(self, curve, x, y, z, t, order=None):
         else:  # pragma: no branch
             self.__coords = (x, y, z, t)
             self.__order = order
+        self.__generator = generator
+        self.__precompute = []
```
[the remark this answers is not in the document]
tomato42: because of `_maybe_precompute()`? the precomputation is delayed until the first time multiplication is performed, so `__generator == True` means "yes, we want to precompute multiplication table", while non empty `__precompute` means we actually did precompute

src/ecdsa/ellipticcurve.py
tomato42: yes, `*self.__coords` is nice syntactic sugar, unfortunately doesn't work on python2, but yeah, could be rewritten

src/ecdsa/ellipticcurve.py
tomato42: hmm, true

src/ecdsa/keys.py
tomato42: yes, as it also allows changing the implementation

src/ecdsa/ellipticcurve.py
tomato42: it is, and I'm almost sure that I wrote the test to do that, must have forgot to include it, will fix

src/ecdsa/ellipticcurve.py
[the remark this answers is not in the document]
t184256: Oh, right. Guess a tri-state value isn't a particularly good idea either.

src/ecdsa/ellipticcurve.py
tomato42: actually we probably want
```python
if r == 0 or r == 2:
    other //= 2
elif r == 1:
    ...
```
as r will be zero about 2/3rds of the time and this is inside a tight loop

t184256:
```
0: 2 comparisons -> 1 comparison
1: 1 comparison -> 2 comparisons
2: 2 comparisons -> 2 comparisons
3: 2 comparisons -> 3-4? comparisons
where comparison = interned const int comparison + jump
```

I'd expect the gains to be real minor even with the unevenness you've mentioned, but if you think it's worth it, why not.

src/ecdsa/ellipticcurve.py
tomato42: done (`test_ed25519_mul_precompute_vs_naf`)

t184256: looks OK

PR #284, opened 2022-02-09
[Docs] Add warning that this is not intended for production use

ijrsvt:
Hey! I was looking for an ECDSA library for my python application and this came up as the first option. It was not super apparent that it is _not intended for production_ use (i.e. it is vulnerable to side channel attacks) and I thought it warranted mentioning this up front. 

I want to be clear, I really appreciate that this library walks through the signing algorithms in an easy-to-understand manner and is much friendlier that staring at C implementations 😄!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ also agree on a shared secret key based on exchanged public keys.
 The keys and signatures are very short, making them easy to handle and
 incorporate into other protocols.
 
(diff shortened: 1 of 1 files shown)


tomato42:
I'd say there are many libraries that are not intended for production use, but I guess putting a link right in the description, warning that this one is too, won't hurt :)

PR #287, opened 2022-03-28
explicit curve encoding fixes

tomato42:
* fix the documentation to list the two valid encoding formats
* Allow the user to control the encoding format of the curve generator when exporting public key with explicit parameters


inikolcev (approved):
Looks good.

PR #289, opened 2022-04-01
better handling for malformed curve parameters

tomato42:
Since explicit curve parameters may not use prime numbers as the field
(see CVE-2022-0778), make sure that our square_root_mod_prime()
handles non-prime p gracefully


beldmit (approved):
LGTM

PR #294, opened 2022-04-12
drop six

xiaoge1001:
do it. https://github.com/tlsfuzzer/python-ecdsa/issues/293

### DIFF
--- a/src/ecdsa/numbertheory.py
+++ b/src/ecdsa/numbertheory.py
@@ -12,8 +12,7 @@
 from __future__ import division
 
 import sys
-from six import integer_types, PY2
-from six.moves import reduce
+from functools import reduce
 
 try:
     xrange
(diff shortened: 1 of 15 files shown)


a-detiste:
@xiaoge1001 thanks for your patch, it has been included on Debian

https://wiki.debian.org/Python3-six-removal

a-detiste:
Actualy, six is not needed for `b("literal")` `b"literal"`also works on Python2;
these lines could already be now merged.

tomato42:
yes change to `b'literal'` is definitely something we can do (and you can see, happens in new code), but there are things like `int_types` that are still needed for py2 compatibility.

a-detiste:
Yes yes... I'm almost done withj my pr "All those moments will be lost in time, like tears in rain."

a-detiste:
Reducing the diff of 80%~90% would make life of downstreams easier.

src/ecdsa/numbertheory.py
```diff
@@ -12,8 +12,7 @@
 from __future__ import division
 
 import sys
-from six import integer_types, PY2
```
cjwatson, line 15: This breaks the `if GMPY2 or GMPY` case: it tries to do `integer_types = tuple(integer_types + (type(mpz(1)),))` and gets a `NameError` as a result.  I suggest adding `integer_types = int,` somewhere up here and reverting the other removals of `integer_types` in this file, as I just did in https://salsa.debian.org/python-team/packages/python-ecdsa/-/commit/3a95d8623d28d73c8c7877e6c3ddd5ee43e33ca4.

PR #298, opened 2022-06-10
Selecting curve by name

tomato42:
Given that curves have two names (the python-ecdsa name and the optional openssl name), finding a curve by name is non-trivial.

Add a helper function for finding curves by name.

Also document the helper function that finds curve by OID

### DIFF
--- a/src/ecdsa/curves.py
+++ b/src/ecdsa/curves.py
@@ -470,6 +470,16 @@ curves = [
 
 
 def find_curve(oid_curve):
+    """Select a curve based on its OID
+
+    :param tuple[int,...] oid_curve: ASN.1 Object Identifier of the
+        curve to return, like ``(1, 2, 840, 10045, 3, 1, 7)`` for ``NIST256p``.
+
+    :raises UnknownCurveError: When the oid doesn't match any of the supported
+        curves
+
+    :rtype: ~ecdsa.curves.Curve
+    """
     for c in curves:
         if c.oid == oid_curve:
             return c
--- a/src/ecdsa/curves.py
+++ b/src/ecdsa/curves.py
@@ -23,6 +23,7 @@ __all__ = [
     "NIST521p",
     "curves",
     "find_curve",
+    "curve_by_name",
     "SECP256k1",
     "BRAINPOOLP160r1",
     "BRAINPOOLP192r1",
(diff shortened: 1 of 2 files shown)


src/ecdsa/curves.py
```diff
@@ -469,10 +470,44 @@ def from_pem(cls, string, valid_encodings=None):
...
     for c in curves:
         if c.oid == oid_curve:
             return c
     raise UnknownCurveError(
         "I don't know about the curve with oid %s."
         "I only know about these: %s" % (oid_curve, [c.name for c in curves])
     )
+
+
+def curve_by_name(name):
```
ZoltanFridrich, line 492: As both functions `find_curve()` and `curve_by_name()` do the same (both find a curve), it would be a good idea to unify their function and argument names. Therefore, I would suggest renaming `find_curve(oid_curve)` to `curve_by_oid(oid)`

src/ecdsa/curves.py
```diff
@@ -469,10 +470,44 @@ def from_pem(cls, string, valid_encodings=None):
...
+    :raises UnknownCurveError: When the name doesn't match any of the supported
+        curves
+
+    :rtype: ~ecdsa.curves.Curve
+    """
+    for c in curves:
+        if name == c.name or (c.openssl_name and name == c.openssl_name):
+            return c
+    raise UnknownCurveError(
+        "Curve with name {0!r} unknown, only curves supported: {1}".format(
```
ZoltanFridrich, line 510: I would also suggest unifying the error message with function `find_curve()`

ZoltanFridrich (approved):
Other then my 2 suggestions, I see no logical errors.
The change looks good.

src/ecdsa/curves.py
tomato42: that's changing API, so I'd rather not do that

src/ecdsa/curves.py
tomato42: What do you mean by unify? It's already using the same exception, and the small difference in how it's phrased is inconsequential.

Yes, find_curve() could have better wording, but without a guideline and enforcement of it in all exceptions, I don't see what we gain by changing it in one place only while making it more confusing for users familiar with old versions.

ZoltanFridrich: If we don't want to change the API or old error messages then leave it as it is.

PR #321, opened 2023-05-26
add support for twisted Brainpool curves

tomato42:
Add support for twisted variants of Brainpool curves


The-Mule (approved):
I did not cross-checked actual parameters of the curves but other than that all looks sane.

tomato42:
thanks for the review!

Issue #111, opened 2019-09-13
Generate usable keypair for OpenSSH?


766F6964:
How can one generate a private-public ECDSA keypair that is in the right 
format to be used for OpenSSH authentication?
The example suggested in the README is not in a format that OpenSSH can understand?
Can someone please show me an example how I can do that?

tomato42:
if `to_pem()` does not work, then that will require support for new format

766F6964:
Yeah, support for OpenSSH would be great.
Currently when trying to authenticate with the .pem file to a remote server I am prompted:
`Load key "priv.key": invalid format`

pmazzini:
@766F6964 Is there any specific curve type you are looking for in here?

PR #323, opened 2023-09-24
export to ssh

pmazzini:
Initial support for exporting keys in the SSH format as requested in #111.
For now it only supports Ed25519 curves but it shouldn't be hard to add additional ones.

### DIFF
--- a/src/ecdsa/keys.py
+++ b/src/ecdsa/keys.py
@@ -7,7 +7,7 @@ from hashlib import sha1
 import os
 from six import PY2, b
 from . import ecdsa, eddsa
-from . import der
+from . import der, ssh
 from . import rfc6979
 from . import ellipticcurve
 from .curves import NIST192p, Curve, Ed25519, Ed448
@@ -612,6 +612,18 @@ class VerifyingKey(object):
             # 0 is the number of unused bits in the
             # bit string
             der.encode_bitstring(point_str, 0),
+        )
+
+    def to_ssh(self):
+        """
+        Convert the public key to the SSH format.
+
+        :return: SSH encoding of the public key
+        :rtype: bytes
+        """
+        return ssh.serialize_public(
+            self.curve.name,
+            self.to_string(),
         )
 
     def verify(
@@ -1280,6 +1292,19 @@ class SigningKey(object):
                 ),
                 der.encode_octet_string(ec_private_key),
             )
+
+    def to_ssh(self):
+        """
+        Convert the private key to the SSH format.
+
+        :return: SSH encoded private key
+        :rtype: bytes
+        """
+        return ssh.serialize_private(
+            self.curve.name,
+            self.verifying_key.to_string(),
+            self.to_string(),
+        )
 
     def get_verifying_key(self):
         """
--- a/src/ecdsa/ssh.py
+++ b/src/ecdsa/ssh.py
@@ -1,0 +1,83 @@
+import binascii
+from . import der
+from ._compat import compat26_str, int_to_bytes
+
+_SSH_ED25519 = b"ssh-ed25519"
+_SK_MAGIC = b"openssh-key-v1\0"
+_NONE = b"none"
+
+
+def _get_key_type(name):
+    if name == "Ed25519":
+        return _SSH_ED25519
+    else:
+        raise ValueError("Unsupported key type")
+
+
+class _Serializer:
+    def __init__(self):
+        self.bytes = b""
+
+    def put_raw(self, val):
+        self.bytes += val
+
+    def put_u32(self, val):
+        self.bytes += int_to_bytes(val, length=4, byteorder="big")
+
+    def put_str(self, val):
+        self.put_u32(len(val))
+        self.bytes += val
+
+    def put_pad(self, blklen=8):
+        padlen = blklen - (len(self.bytes) % blklen)
+        self.put_raw(bytearray(range(1, 1 + padlen)))
+
+    def encode(self):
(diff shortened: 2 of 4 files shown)


tomato42:
IIRC, openssh supports only 4 or 5 curves (P-256, P-384, P-521, Ed25519, Ed448), I see no reason not to support all of them...

src/ecdsa/keys.py
```diff
@@ -614,6 +614,18 @@ def to_der(
             der.encode_bitstring(point_str, 0),
         )
 
+    def to_ssh(self):
```
tomato42, line 617: the public format normally includes a comment too, shouldn't we have that as an optional function argument?

src/ecdsa/keys.py
```diff
@@ -1281,6 +1293,19 @@ def to_der(
                 der.encode_octet_string(ec_private_key),
             )
 
+    def to_ssh(self):
```
tomato42, line 1296: what about `from_ssh`?

src/ecdsa/ssh.py
```diff
@@ -0,0 +1,80 @@
...
+
+class _Serializer:
+    def __init__(self):
+        self.bytes = b""
+
+    def put_raw(self, val):
+        self.bytes += val
+
+    def put_u32(self, val):
+        self.bytes += val.to_bytes(length=4, byteorder="big")
```
tomato42, line 24: I don't think this will be portable to old python versions...

tomato42 (changes requested):
test coverage missing

beldmit:
> IIRC, openssh supports only 4 or 5 curves (P-256, P-384, P-521, Ed25519, Ed448), I see no reason not to support all of them...

Not sure about Ed448

src/ecdsa/keys.py
pmazzini: I am focusing only on exporting for now, is that an issue?

pmazzini:
> test coverage missing

I am happy to add some tests if you think the code is looking ok-ish.

pmazzini:
> Not sure about Ed448

Yeah, there is no support for Ed448 yet.

src/ecdsa/ssh.py
pmazzini: Ok, I'll take a look at this.

src/ecdsa/keys.py
pmazzini: That can be added afterwards if really needed. From other implementations it does not look to be something it is being used:

https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/serialization/ssh.py#L745

src/ecdsa/ssh.py
pmazzini: Actually, what do you mean with old Python versions? Are you referring to end of life ones?

https://devguide.python.org/versions/

tomato42: Yes, we explicitly support python 2.6. Just because upstream doesn't support it, doesn't mean there aren't supported distributions of it.

src/ecdsa/keys.py
tomato42: every other key operation has support for both import and export. And for test coverage it would be good to not only test against static vectors, but also to verify that a random key can be imported and exported.

src/ecdsa/keys.py
tomato42: just an idea; I'm not insisting on it

tomato42:
> > test coverage missing
> 
> I am happy to add some tests if you think the code is looking ok-ish.

yes, the code looks ok-ish, but python 2.6 compat and test coverage is mandatory, I won't merge the code otherwise.

pmazzini:
Added unit tests. CI for Py 2.x was working fine.

tomato42:
you will need to use methods from https://github.com/tlsfuzzer/python-ecdsa/blob/master/src/ecdsa/_compat.py for py2 compat

tomato42 (changes requested):
CI failures are relevant

pmazzini:
Switched to compat int_to_bytes.

tomato42 (changes requested):
CI failures on Python 2 still.

Also, please note the  code checks run: https://github.com/tlsfuzzer/python-ecdsa/actions/runs/6374029084/job/17310264458?pr=323

src/ecdsa/ssh.py
```diff
@@ -0,0 +1,85 @@
...
+    ciphername = kdfname = _NONE
+    main.put_str(ciphername)
+    main.put_str(kdfname)
+    nokdf = 0
+    main.put_u32(nokdf)
+    nkeys = 1
+    main.put_u32(nkeys)
+    main.put_str(spub.tobytes())
+    main.put_str(spriv.tobytes())
+    return main.topem()
```
tomato42, line 85: why `der.topem()` can't be used?

pmazzini:
Switch to using `der.topem()` and fix it while on it.
Use a line break every 76 bytes: https://docs.python.org/3/library/base64.html#base64.encodebytes

pmazzini:
Use `compat26_str` in `der.topem()`.

pmazzini:
More `compat26_str`.

pmazzini:
Any blocker for merging? :)

tomato42 (approved):

tomato42:
> Any blocker for merging? :)

yes, my time to do a final review :)

thanks for the PR!

PR #324, opened 2023-09-30
Remove unnecessary adds from mul_add

vojtechsu:
For the computation of mul_add, four adds are being precomputed (A+B, A-B, -A+B, -A-B). Two of them are unnecessary as A+B and -A-B only differ in the sign of the y-coordinate. The same holds for -A+B and A-B.


tomato42:
good point, thanks

sorry, I mixed up branches on push and closed this one unintentionally, I've restored the patch in PR #326

PR #339, opened 2024-04-24
trim more usage of six.b()

a-detiste:
here is a new patch to reduce patching in Debian while keeping here the package Python2.6 compatible

greetings


tomato42 (approved):
ooh, nice! Thank you!

PR #344, opened 2024-08-06
handle non-prime order curves more gracefully

### LINKED ISSUE #343: 3 Bugs on general ECC calculation
Here I construct a ECC $y^2 = x^3 + x + 1 mod 23$ . And starting from point $(9, 7)$. Here are three bugs I found. Any body who is intreated can reproduce the bugs easily.

```
from ecdsa.ellipticcurve import CurveFp, Point, PointJacobi


if __name__ == '__main__':
    curve = CurveFp(23, 1, 1)
    G = PointJacobi(curve, 9, 7, 1)
    G_affine = G.to_affine()
    print('14*G+G==15*G:', 14*G + G == 15*G)  # Bug1: 14*G + G != 15*G !!!
    print('14*G+G==15*G:', 14*G_affine + G_affine == 15*G_affine)
    G_27 = 27*G  # Bug2: This results in a negative y coordinate.
    print('G_27: (%d, %d)' % (G_27.x(), G_27.y()))
    G_affine_27 = 27*G_affine  # Bug3: This raises an exception.
    print('G_affine_27: (%d, %d)' % (G_affine_27.x(), G_affine_27.y()))
```

tomato42:
when the order of the curve is not a prime, then point doubling can return INFINITY, this will cause some negative values not to be reduced modulo curve p; fix this

fixes #343

### DIFF
--- a/src/ecdsa/ellipticcurve.py
+++ b/src/ecdsa/ellipticcurve.py
@@ -633,7 +633,7 @@ class PointJacobi(AbstractPoint):
         """
         x1, y1, z1 = self.__coords
         if other is INFINITY:
-            return not y1 or not z1
+            return (not x1 and not y1) or not z1
         if isinstance(other, Point):
             x2, y2, z2 = other.x(), other.y(), 1
         elif isinstance(other, PointJacobi):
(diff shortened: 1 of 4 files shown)


ZoltanFridrich (approved):
I haven't spotted any mistake. Looks OK to me.

PR #354, opened 2025-02-20
Fix for CVE-2024-23342.

atul-bhardwaj:
Following fix saves against minerva attack.


tomato42:
Sorry, this is fixing nothing. And as I wrote multiple times, it's not possible to fix without changing the implementation to C

PR #356, opened 2025-03-06
add support for parsing implicit DER tags

tomato42:
While not needed for ECDSA/ECDH PKCS#8 files, it looks like it will be necessary for the PQC algorithms, so provide that too, so we can do some code reuse.

### DIFF
--- a/src/ecdsa/test_der.py
+++ b/src/ecdsa/test_der.py
@@ -22,6 +22,7 @@ from .der import (
     remove_object,
     encode_oid,
     remove_constructed,
+    remove_implicit,
     remove_octet_string,
     remove_sequence,
 )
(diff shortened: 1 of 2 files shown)


src/ecdsa/test_der.py
```diff
@@ -396,6 +398,128 @@ def test_with_malformed_tag(self):
...
+    @classmethod
+    def setUpClass(cls):
+        cls.exp_tag = 6
+        cls.exp_data = b"\x0a\x0b"
+        # data with application tag class
+        cls.data_application = b"\x46\x02\x0a\x0b"
+        # data with context-specific tag class
+        cls.data_context_specific = b"\x86\x02\x0a\x0b"
+        # data with private tag class
+        cls.data_private = b"\xc6\x02\x0a\x0b"
```
rjrelyea, line 411: We probably should have a test for length > 128 to make sure the llen encoding/decoding is happening correctly

rjrelyea (approved):
I have a suggestion, which you can add before merge or ignore.

src/ecdsa/test_der.py
tomato42: this is reusing common code for encoding and decoding length, so I'm quite sure that it will handle it correctly

PR #382, opened 2026-09-07
keys: VerifyingKey.precompute() works for keys built from bytes, PEM or DER

fametrano:
On the Weierstrass curves, `VerifyingKey.precompute()` raises `AssertionError` for every key built with `from_string`, `from_pem` or `from_der`. On 0.19.2:

| built with | `precompute()` |
|---|---|
| `SigningKey.from_secret_exponent(...).verifying_key` | works |
| `VerifyingKey.from_string(raw)` / `from_string(compressed)` | `AssertionError` |
| `VerifyingKey.from_pem(...)` / `from_der(...)` | `AssertionError` |

Cause: `from_string` calls `PointJacobi.from_bytes` without `order`, so `point.order()` is `None`; `precompute()` then hits `assert order` in `_maybe_precompute`. `from_pem` and `from_der` go through `from_string`. Keys from a `SigningKey` or from signature recovery carry the order and are not affected; EdDSA keys take a separate branch.

Fix: pass `order=curve.order` (one line). Test: `precompute()` then `verify()` after each of the four constructions; it fails on master and passes with the fix. Full suite: 2040 passed.

The docstring says to call `precompute()` when verifying hundreds of signatures under one key, which is exactly the caller who has the key as bytes, PEM or DER. Measured on secp256k1, a verification drops from about 1057 µs to 501 µs with the table.


fametrano:
@tomato42 the two workflow runs on this PR (GitHub CI and CodeQL) have been waiting for approval since I opened it, so CI has not run here yet — it is my first PR to this repo. Could you approve them?

Re-checked today on the current tip `02c9cb7`, which is up to date with master: the full suite gives 2040 passed, 5 skipped.

